PR #2, opened 2022-11-16
Install lando / switch to postgres

mccalluc:
I copied the lando config files from the template, rather than running lando init.


mccalluc:
wrong direction. Closing.

PR #4, opened 2022-11-16
Add rubocop and rspec

mccalluc:
Use the rails generator for rspec, but copy the rubocop.yml from the template; Make corrections required by rubocop.


mccalluc:
wrong direction. Closing.

PR #15, opened 2022-11-18
explain credentials

mccalluc:
I'm unclear on this: Do we use both envvars and config files?

If this seems good, would it be ok to suggest it upstream in the rails-template?


config/credentials.yml.enc
```diff
@@ -1 +1,4 @@
-Q0NZBPwCK18TE0EO8hQD4Xbi65F1s2lvBFPv9LvgqbfOEKkPEoWsVIYpQW3BD6SpwlJdbvnMoHhOrkz0ugeHzkMp/20XcwER0yvkMbEWg5fsgmNFvsBaOkF5Hct1oqpSOYVm3ZEFlSDB7YX1YXAqHfwpEqY11q4p+SHUFQM9d4MRB13XNFWwAbFGflnb2oNU/w4aDFNw6ovc0+FsaYxve25oJdU0REWL6Tt2ONaNV842m+nMAcXd2pqIPqfQqS7nOVhTMHbsCrRfTQFvCwV19kE2f6sE8qTUkR9FXHij2bJlDrSzrfOXcXVr/do05U/ntV21ESuTs8vNn8OjBJ5T3TMncWyFrPkRq4IDW/hF2Mny03UM8/nptTa5DTYP8/JY67bk4z/byVzEOOamOmI5uu6UUIVENIaV3y60--NOSBNFn50kH+PZtZ--/MSSStmr1n+64mgZW0L7aQ==
\ No newline at end of file
+# The default Rails way to manage secrets is with this credentials file
+# that *is* checked in, and a master.key that is gitignored.
+#
+# However, at PUL we instead use environment variables, and gitignored config files.
```
carolyncole, line 4: I would change this to something along the lines of
```suggestion
# However, at PUL we instead use environment variables configured in princeton_ansible, which are reference in config files via the rails ENV[:blah].
```

carolyncole (changes requested):
A tweak to more closely match what we do in PUL

PR #19, opened 2022-11-30
Login with plain credentials

mccalluc:
Draft: This is throw-away work that moves us towards collecting the MF username and password with a plain Rails form... but by January an integration between CAS and MF might be available... so I'll leave it as a draft.


mccalluc:
Not the direction we're going.

Issue #22, opened 2022-12-22
Project data model


matthewjchandler:
### Notes

- _This issue is about defining a Project for the purposes of early development_
  - Many of the policies surrounding Projects and required metadata are yet to be fleshed out, but we can sketch out the skeletal structure of Projects in our preliminary data models, following the notes below...
- The lynchpin for all roles, metadata, and data in TigerData is the Project
  - Every file in TigerData must belong to a Project
  - Every Project must have defined roles
  - Every Project will have its own metadata, including the roles and other descriptive information
  - Any optional metadata assigned beyond the required minimal metadata for data files will be managed at the Project level
- There are 3 main roles for Projects (with more roles coming later, along with evolving rules for each role):
  - Data Sponsor (required for every Project; singular for now):
    - The person who bears primary responsibility for a Project
    - Must be PI-eligible or otherwise approved by Steering Committee policy
    - Has all privileges available within a given Project
  - Data Manager (required for every Project; singular for now):
    - The person who does the day-to-day data management
    - Must have completed the training offered by PRDS
    - Must be assigned by the Data Sponsor
    - Can be the same person as the Data Sponsor
    - Has privileges to assign/edit metadata and perform/request transfer operations, as well as all Data User privileges.
  - Data User (optional and multiple):
    - The persons who have read access to data and metadata within a Project
    - Must be Princeton affiliates (for now)
    - Assigned by the Data Sponsor
- There are 2 main types of metadata for a Project provided by the Data Sponsor (possibly expanding, with evolving rules):
  - Roles: as above
  - Description:
    - Project title
    - Short memo on the purpose of the Project
    - Funding source(s)
    - Start and end dates (can be estimated)
- As we establish the infrastructure for TigerData, Projects will have associated storage provisions and data files
  - We will need procedures for requesting increases to Project storage allocations and mounting storage locations for Data User access
  - How exactly Projects will "contain" files is an open question
- Projects will be the conduit through which metadata is assigned (usually in batch) to data files
  - There will soon be a policy established for minimal metadata for all data in TigerData
  - Soon after, there will be a framework for applying discipline-specific metadata schemas to data under certain Projects
  - And eventually, there will be a way for Data Sponsors and Data Managers to apply their own bespoke metadata to their Projects
- We should assume there will be need to split and merge Projects, to arrange Projects and Subprojects in a hierarchy, and to define other relations among different Projects

### Related User Journey

[...]

PR #23, opened 2023-01-04
Initial ER diagram

mccalluc:
I believe this captures all the details from #22 that this kind of diagram can capture.


mccalluc:
Feedback: The second diagram is more what we want, with the addition of a `Role` entity. The precise roles that we'll need is still in flux, and should not be hard-coded in the RDB structure. Back to draft.

bess (approved):

PR #24, opened 2023-01-10
Propose add `AllowedRole`; remove booleans

mccalluc:
Hope to get feedback. This may be too general.


mccalluc:
January 11, 2023 meeting: Matt indicated there would be more roles. Up to us how this is implemented behind the scenes, but I'm inclined to take the more general approach outlined here.

bess (approved):

PR #32, opened 2023-02-08
Fresh Rails installation

### LINKED ISSUE #29: Regenerate Rails application
Configure the way we want a Rails application configured (including CI).

### LINKED ISSUE #30: Add the tiger data logo to the Application Header
Add the logo to the application:  https://drive.google.com/drive/folders/1wBV1X-5aMxl-rn_aOXiSGByQBDSbL8qv

mccalluc:
- towards #29: First attempt to move to postgres from the rails default mysql did not work: Needs to be done, but I'll do that in a follow-up, since this much works.
- fix #30
- new todo: #33
- new todo: #34
- Unsure about `config/deploy.rb`: There is one note in [rails-template](https://github.com/pulibrary/rails-template#capistrano) about modifying the autogenerated file, but more changes may be needed.
- Unsure about `config/initializers/assets.rb`: Copy-and-paste version references yarn... but it also references node_modules, which is not in use here. More work may be needed to get JS working.

An earlier meeting clarified that https://github.com/pulibrary/rails-template should not be used as a template: That is, we should not start a project by copy-and-pasting those files, and then making changes as needed. Instead, the consensus was that new PUL projects should start by running the rails generator, and then the rails-template README can be followed for details. For some configurations, I've copied from `pdc_describe`, because it is up to date, and reflects the latest preferences of our team.

That same meeting considered whether it would be worth having a more automated process for getting a new project started, and the answer was no: We don't have new projects frequently, but the dependencies do change rapidly, so doing it by hand seemed like the best course... but having attempted it, I do think there is a lot to be said for a dockerfile or something similar that would make each step of the process explicit and easily re-runnable

I've also added eslint, based on pdc_describe, which is not in rails-template, and I've added a route for the homepage, with a down-scaled copy of the logo, and a controller test.


bess (approved):
🎉

PR #39, opened 2023-02-15
Migrate to vite; Resolve ESLint warnings

### LINKED ISSUE #33: Fix eslint errors in rails generated JS
Right now there are ignores in `.eslintrc.yaml`: These should be removed.

### LINKED ISSUE #38: Migrate JS to Vite
Before worrying about the few ESlint errors:
- #33

... we should migrate to Vite. The JS files that came with the default rails build probably aren't needed, or won't be in the same place with vite.

mccalluc:
- Fix #33
- Fix #38
- Simplify ESLint invocation

Spockets functionality overlaps with Vite: Not sure if it should be removed too. Filed follow-up issue:
- See #40


mccalluc:
resolving merge conflicts... back to draft.

carolyncole:
@mccalluc does this run for you?  When I run it locally I get the following error:
Screenshot 2023-02-17 at 9 19 15 AM

mccalluc:
Let me update the branch and check...

mccalluc:
hmm: after merging main, updating gems to add capistrano, and restarting, the homepage still works... but there still could be something weird on my side that's helping it to work.

There aren't any system specs right now, so let me add that on this branch, and that may help determine if there is something else that is missing.

mccalluc:
@carolyncole -- Thanks: I must have had extra JS packages locally which weren't fully specified in the config. A new system spec reproduced your exact error, and with these changes it is passing. Please try again.

- (I've filed #48, but I don't think that needs to block this PR.)

carolyncole (approved):
Runs for me locally now (After running `yarn install`)

PR #42, opened 2023-02-16
Copy and paste write_version task as suggested by rails-template

### LINKED ISSUE #31: Is version.txt needed?
Trying to follow the instructions on `rails-template`:
> Add the `write-version` task to give your project a public page at `/version.txt` with the branch name, commit SHA, and release date of the current deployment.

There is [no other reference](https://github.com/pulibrary/rails-template/search?q=write-version) to `write-version` in that repo, and I also can't find anything in `pdc-describe`.

If this is is something tiger-data should have, please clarify or point to an example. If it's not needed this can be closed. Perhaps remove the mention from `rails-template`.

mccalluc:
- Fix #31

This is part of the setup instructions in rails-template, but it is in place only [five PUL projects](https://github.com/search?q=org%3Apulibrary+write_version&type=code). I don't care strongly about this -- I think it would also be ok to close this PR and the corresponding issue without merging.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -8,3 +8,13 @@ set :repo_url, "https://github.com/pulibrary/tiger-data-app.git"
 set :deploy_to, "/opt/tigerdata"
 
 set :branch, ENV["BRANCH"] || "main"
+
+desc "Write the current version to public/version.txt"
+task :write_version do
+  on roles(:app), in: :sequence do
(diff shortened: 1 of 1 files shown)


carolyncole (approved):
This creates a file like: https://dpul.princeton.edu/version.txt

I like it, but it is different from what we usually utilize

PR #45, opened 2023-02-16
Middleware/API architecture proposal

### LINKED ISSUE #10: Clarify architecture for Mediaflux client
Team not ready to make a decision right now. One possibility:

- Client: generic, few public methods, parses xml response, makes error responses into exceptions
- Wrapper: multiple public methods, assembles XML, pulls out data from response
- Models: Decoupled from MF details

Or would just two layers be sufficient, with wrapper responsibilities moved either up or down?

mccalluc:
- Fix #10

Matt has emphasized
- That everything the UI offers should also be available in the API
- That there is some kind of "Middle Layer" between the UI/API "Presentation Layer" and the "Data Management System"

This is an attempt to pin down an implementation, and I'll link to this from the agenda to make sure there can be a consensus around it. The proposal is to have an `ApiMiddleware` class that will contain all references to either the database or mediaflux:
```
  # The database and Mediaflux should only be accessed through this class.
  # Since the respective responsibilities of the the two systems have not
  # been determined, we want all references to both in just one place.
  #
  # As the middleware grows to support the needs of the UI,
  # we need to be sure to add corresponding API routes.
```

```mermaid
flowchart LR
  UI --> ApiMiddleware --> Postgres
  API --> ApiMiddleware --> Mediaflux
```

A stricter alternative would be to require that the UI use the API: This would guarantee that the UI doesn't get ahead of the API. This might make sense if we had a separate microservice for the API, but when it's all on one server, it seems gratuitous.

```mermaid
flowchart LR
  UI --> API --> ApiMiddleware --> Postgres
  ApiMiddleware --> Mediaflux
```

A looser alternative would be to let us use Rails ActiveRecord where ever we want it, and limit the middleware to just a wrapper around Mediaflux.
```mermaid
flowchart LR
  UI --> ApiMiddleware
  API --> ApiMiddleware --> Mediaflux
  UI --> Postgres
  API --> Postgres
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ coverage
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+
+# Byebug
+.byebug_history
--- a/app/controllers/api/v0/projects_controller.rb
+++ b/app/controllers/api/v0/projects_controller.rb
@@ -2,7 +2,9 @@
 module Api
     module V0
         class ProjectsController < ApplicationController
-            def index; end
+            def index
+                render json: "{}"
+            end
         end
     end
 end
--- a/app/controllers/api/v0/projects_controller.rb
+++ b/app/controllers/api/v0/projects_controller.rb
@@ -3,7 +3,8 @@ module Api
   module V0
     class ProjectsController < ApplicationController
       def index
-        render json: ApiMiddleware.new.projects.to_json
+        projects = ApiMiddleware.new.projects
+        render json: projects.to_json
       end
     end
   end
--- a/app/controllers/api/v0/projects_controller.rb
+++ b/app/controllers/api/v0/projects_controller.rb
@@ -1,11 +1,10 @@
 # frozen_string_literal: true
 module Api
-    module V0
-        class ProjectsController < ApplicationController
-            def index
-                render json: ApiMiddleware.new().projects.to_json
-            end
-        end
+  module V0
+    class ProjectsController < ApplicationController
+      def index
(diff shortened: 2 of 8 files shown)


escowles:
I thought I'd share a couple of thoughts and questions about how this would work:

1. To what extent would Rails controllers fulfill the role of the API (i.e., the views would constitute the UI)? Would scrupulously creating controller endpoints for any data that controllers or views/helpers fill the gaps?
2. I've seen token-based auth folded into controllers so browsers and automated clients use the same endpoints, but I think it's more common to have different endpoints.
3. One of the ramifications of having a significant external command-line/API customer base that Matt, Kate, and I discussed recently is that any backwards-incompatible changes would potentially break their integrations. Maybe this argues against considering Rails controllers the API?
4. What are the best practices for having a separate, versioned API?
5. I'm not sure that having a "Middle Layer" between the UI/API and the backend is a requirement, or if it's one possible approach to making sure that users don't have unmediated access to MediaFlux just because they are using a command-line tool or API integration. Basically, maybe this app is the "Middle Layer"?

mccalluc:
Notes from this morning:

> Bess: would like to have a separate design session, with concrete use cases.
> Esme’s concerns about upgrade path are important
> Kate: Talking with Matt, would like to get concrete usecases
> Hector: Would like to start as simply as possible. We probably won’t have enough information for the next few months. If we start wrong it could be a pain to maintain
> Bess: We don’t have solid requirements right now… we’ll probably need to rewrite it.
> Carolyn: Controller is the api
> Close this PR: Make a new one where it’s in the controller

PR #51, opened 2023-02-17
Simpler architecture proposal

### LINKED ISSUE #10: Clarify architecture for Mediaflux client

mccalluc:
- Replaces #45 
- Fix #10

Without trying to prescribe how auth will work, or what the relational data or mediaflux data models are, or what will even be handled by mediaflux and what will be handled by postgres, or whether the API will be GraphQL or something simpler, this PR attempts to clarify the software architecture.

- UI and API will be independent: It is up to the developers to ensure that the functionality of the API matches the UI.
- We'll start a `MediafluxWrapper` class to contain all code related to Mediaflux. Eventually, credentials will probably be passed to the constructor.


app/models/mediaflux_wrapper.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+class MediafluxWrapper
+  # All calls to Mediaflux should be made through this class.
+  # TODO: Figure out authentication and authorization.
+  # TODO: Figure out what data Mediaflux and Postgres are responsible for.
+  # (Projects may eventually be handled by Postgres, but I need something for the example.)
```
jrgriffiniii, line 6: 👍

spec/controllers/api_v0_projects_controller_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.describe Api::V0::ProjectsController do
+  render_views
+
+  it "renders the index page" do
+    get :index, params: { format: :json }
+    expect(response.body).to include("[{")
```
jrgriffiniii, line 9: 👍

jrgriffiniii (approved):

Issue #21, opened 2022-12-22
User login and authentication


matthewjchandler:
### User Story

_As a TigerData User, I need to log into the web portal, so that I can review my roles and responsibilities._

### Notes

- Methods of authentication will evolve as the Presentation Layer matures, but for early development, we can use a Rails strategy (CAS)
- Users will all be known Princeton affiliates to start, but eventually, we will need login/authentication methods for external collaborators
- Not all Princeton affiliates are TigerData users; we will need a layer of verification beyond mere affiliation with the University

### Related User Journey

![TigerData_PresentationLayer_UserJourneys_v0 3_2022-12-21_pg03](https://user-images.githubusercontent.com/66128387/209208296-96cbc4ff-ce70-46a8-ae8e-f29e17176f8b.png)
(Page 3 in [https://drive.google.com/file/d/17OW4p5a1zdc0SU46mrd1LDLI0k2Nz_ua/view?usp=share_link](url))

### Completion Criteria

- [x] #64
- [ ] Welcome message top-center on the home page: "Welcome to the TigerData user portal. Please log in."
- [ ] The words "log in" on the welcome message on link to the preliminary authentication method (CAS)
- [ ] Preliminary authentication method verifies that the person logging in is a current Princeton affiliate
- [x] #61
- [ ] Upon successful authentication as a TigerData user, user is taken to a placeholder dashboard page (e.g., tigerdata.princeton.edu/dashboard)
- [ ] Upon successful authentication as a Princeton affiliate only, user is taken to a placeholder "coming soon" page
- [ ] Upon failed authentication, user is taken back to the login page

PR #63, opened 2023-02-20
Add CAS auth

mccalluc:
- Towards #21

Draft: Hitting an error:
```
undefined method `session_path'
```

A lot of this is just following steps from the documentation, so I'll try starting over with a fresh branch, in case I missed something.


mccalluc:
Made a fresh clean branch, but hit the same error.

PR #78, opened 2023-02-22
Demonstrate how jQuery unit tests could work

### LINKED ISSUE #62: Demonstrate JS/jQuery unit testing
While our overall test coverage is good, JS is only covered by system tests. It would be good to at least have a handle on how we would write unit tests for JS.

jQuery is a little awkward to test because it's often going to depend on particular HTML structures that come from the ERB, so we would need to recreate these structures in the JS test, and these will need to be kept in sync.

Jest is a popular JS testing framework: [Their docs](https://jestjs.io/docs/tutorial-jquery) provide an example of testing jQuery code.

mccalluc:
- Fix #62


mccalluc:
Turns out that Jest may not be a good fit with Vite. I'm going to start a new PR and try a project called [vitest](https://vitest.dev).

PR #85, opened 2023-02-27
Demonstrate JS testing

### LINKED ISSUE #62: Demonstrate JS/jQuery unit testing

mccalluc:
I had started this before the meeting where we closed
- #62

We don't really need it for anything right now: I think it would be fine to close this unmerged... but I did learn something by implementing it, and it might be useful in the future. Your call.

### DIFF
--- a/app/javascript/entrypoints/helper.js
+++ b/app/javascript/entrypoints/helper.js
@@ -1,3 +1,4 @@
+// eslint-disable-next-line import/prefer-default-export
 export function setTargetHtml(event, html) {
-    $(event.target).html(html)
+  $(event.target).html(html);
 }
--- a/app/javascript/entrypoints/helper.js
+++ b/app/javascript/entrypoints/helper.js
@@ -1,0 +1,3 @@
+export function setTargetHtml(event, html) {
+    $(event.target).html(html)
+}
--- a/app/javascript/tests/helper.test.js
+++ b/app/javascript/tests/helper.test.js
@@ -1,7 +1,10 @@
 // @vitest-environment jsdom
 
-import { assert, describe, expect, it } from 'vitest'
-import { setTargetHtml } from '../entrypoints/helper';
+import {
+  // assert,
+  describe, expect, it,
+} from 'vitest';
+// import { setTargetHtml } from '../entrypoints/helper';
 
 describe('helper', () => {
   it('has a window', () => {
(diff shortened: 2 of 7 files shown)


mccalluc:
Ruby test fails because Geckodriver installation doesn't work... I don't think changes in this PR would have touched that. Will rerun.

mccalluc:
I had started this before the meeting where we closed
- #62

We don't really need it for anything right now: I think it would be fine to close this unmerged... but I did learn something by implementing it, and it might be useful in the future. Your call.

app/javascript/entrypoints/helper.js
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/prefer-default-export
+export function setTargetHtml(event, html) {
```
jrgriffiniii, line 2: 👍 I am in no way opposed to this or critical, but was this separated in order to decouple from jQuery? I can follow this pattern if it is preferred, but I just wish to be consistent should I implement JS elsewhere.

app/javascript/tests/helper.test.js
```diff
@@ -0,0 +1,21 @@
...
+describe('helper', () => {
+  it('has a window', () => {
+    expect(typeof window).toBe('object');
+  });
+
+  it('can test jQuery', () => {
+    const dom = new JSDOM('<!DOCTYPE html><h1>Hello world</h1></html>');
+    const h1 = dom.window.document.querySelector('h1');
+    setTargetHtml({ target: h1 }, 'Goodbye');
+    expect(h1.innerHTML).toEqual('Goodbye');
```
jrgriffiniii, line 19: 🎉

jrgriffiniii (approved):
Thank you very much for this! I did have one question, however I can raise this during the check-in meeting.

app/javascript/entrypoints/helper.js
mccalluc: Fuller discussion in standup. For reference, it's not about decoupling jQuery, it's about moving testable functions/classes out of application.js and into modules that just export with no side effects.

Issue #58, opened 2023-02-20
User friendly view of a Mediaflux schema


mccalluc:
Beyond just rendering the schema, would we like to dynamically create HTML forms given a MF schema? Is there a usecase for _just_ printing out the schema, apart from a data entry form?

PR #94, opened 2023-02-28
Simple util function for rendering schema definition XML

mccalluc:
- Towards #58

Eventually, we'll want want to call mediaflux to list the asset types and get their definitions, but once we have the definition, we'll need to be able to show it. This is one relatively simple approach. Feedback welcome:

- Invest time now in a better rendering? (Perhaps combine attribute name values; Real html instead of pre-formatted; ...)
- Is there a better place for the code than `lib/`?

### DIFF
--- a/lib/xml_utilities.rb
+++ b/lib/xml_utilities.rb
@@ -1,18 +1,32 @@
+require 'yaml'
+
 # frozen_string_literal: true
 TYPENAMES = { 1 => "element", 2 => "attribute", 3 => "text", 4 => "cdata", 8 => "comment" }.freeze
 
-def xml_to_hash(document)
+def xml_node_to_hash(node)
   # Based on https://stackoverflow.com/a/10144623
-
-  node = document.root
-
-  { kind: TYPENAMES[node.node_type], name: node.name }.tap do |h|
+  {}.tap do |h|
+    h[:name] = node.name
+    node_type_name = TYPENAMES[node.node_type]
+    if node_type_name != "element"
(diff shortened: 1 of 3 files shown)


lib/xml_utilities.rb
```diff
@@ -0,0 +1,34 @@
...
+  h
+end
+
+def xml_doc_to_hash(document)
+  xml_node_to_hash(document.root)
+end
+
+def xml_doc_to_html(document)
+  as_hash = xml_doc_to_hash(document)
+  as_yaml = YAML.dump(as_hash).gsub(/\A---\n/, "")
```
jrgriffiniii, line 32: 👍

spec/lib/xml_utilities_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  :attributes:
+  - :name: name
+    :text: my_string
+  - :name: type
+    :text: string
+</pre>"
+)
+  end
+
+  # If we want more tests or longer tests, they should be checked in as fixtures.
```
jrgriffiniii, line 28: 👍

jrgriffiniii (approved):

jrgriffiniii (approved):
Please note that there is some test coverage missing (please see https://output.circle-artifacts.com/output/job/35c67965-fca3-4dc1-94fa-63f372063e7f/artifacts/0/coverage/index.html#3d8311a6d4930456d5ecf8e9d63e681de1773243), however this should not block this being merged.

Issue #98, opened 2023-03-01
Demonstrate connectivity to MediaFlux in the UI


kelynch:
List the files in a namespace from MediaFlux into the UI.  The namespace can be hard-coded at this time, for the purposes of demonstrating our ability for the Rails application to communicate with the MediaFlux server.

After you login you will see a list of items from a path in mediaflux

- [x] A logged in user: On the home page a list of collection assets is shown from in the base path
- [x] A non logged in user should see nothing
- [x] No test should be created

Utilize the code in [Experiments](https://github.com/pulibrary/tiger-data-experiments/blob/main/simple-nokogiri/list_collections.rb#L8-L28)

kelynch:
Ideally we'd love to demo this during the stakeholder retreat (March 13).

jrgriffiniii:
Specifying the `namespace` arguments when listing the collections seems to be required. While I have been attempting to adjust the `size` argument, this still triggers server response timeout errors.

jrgriffiniii:
I apologize, I have progressed with https://github.com/pulibrary/tiger-data-app/compare/jrgriffiniii-mflux-collection-query?expand=1 thus far. I am sorry for the absence of test coverage, I can look to address this following a meeting this afternoon.

PR #115, opened 2023-03-09
Proposed Mediaflux Client restructuring

jrgriffiniii:
This is to propose a separation of concerns between those classes which are responsible for interfacing with the Mediaflux server API, and those classes which model resources within the Mediaflux server as first-order entities (e. g. Collections).

As these are merely proposals, I did not seek to abstract what has been implemented for Collections into abstract base classes which might be required for Assets (amongst many, many other shortcomings). This is primarily to please request feedback regarding this approach, and any critical or negative feedback is most welcome if this is not sufficient.

Advances #98


spec/support/stub_mediaflux.rb
```diff
@@ -0,0 +1,75 @@
...
+      ).to_return(status: 200, body: collection_list_response_body)
+
+      # Stubbing the system.logon request
+      stub_request(:post, "http://0.0.0.0:8888/__mflux_svc__").with(
+        body: /<service name="system.logon">/
+      ).to_return(status: 200, body: logon_response_body)
+    end
+  end
+
+  config.before(:each) do |ex|
```
carolyncole, line 70: @jrgriffiniii this is just a bit confusing to me.  Should this be an after?  Otherwise you are enabling webmock and then immediately disabling it.

carolyncole (changes requested):
I'm happy for the refactor.  Just a couple questions/ comments...

jrgriffiniii:
Deprecated by https://github.com/pulibrary/tiger-data-app/pull/126

Issue #116, opened 2023-03-09
Revisit demo MediaFlux code, post-retreat


kelynch:
After the RDSS retreat day (March 13, 2023), revisit [this code](https://github.com/pulibrary/tiger-data-app/pull/114), [this code](https://github.com/pulibrary/tiger-data-app/pull/119), [this code](https://github.com/pulibrary/tiger-data-app/pull/120), and [this code](https://github.com/pulibrary/tiger-data-app/pull/118), and either remove or refactor it to pass CI/work with our agreed-upon programmatic approach sustainably.

Acceptance criteria:
- [ ] Add pagination to classes https://github.com/pulibrary/tiger-data-app/tree/main/app/models/mediaflux
- [ ] Add additional classes to https://github.com/pulibrary/tiger-data-app/tree/main/app/models/mediaflux
- [ ] Edit the code and update to utilize helper classes in https://github.com/pulibrary/tiger-data-app/tree/main/app/models/mediaflux

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/116?planning-poker) with Zenhub @bess @jrgriffiniii

PR #130, opened 2023-04-07
Utilizing Mediaflux HTTP pattern in media_flux_client

carolyncole:
Adding classes to run the commands

refs #116

### DIFF
--- a/app/controllers/welcome_controller.rb
+++ b/app/controllers/welcome_controller.rb
@@ -4,18 +4,11 @@ class WelcomeController < ApplicationController
   def index
     return if current_user.nil?
 
-    start = params[:start].nil? ? 1 : params[:start].to_i
-    @mf_version = media_flux.version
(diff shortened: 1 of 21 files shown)


carolyncole:
1. Merge @hectorcorrea's PR 
1. Break this PR into sections 
    1. to user linked to session so we do not login every time
    1. Add New Classes that exists in PR and merge mediaflux-client
    1. Session object get instantiated and the new Classes from @hectorcorrea's PR call the Mediaflus::Http classes instead of the mediaflux-client

td-meta1 is slower:  No HTOP, not ubuntu
1. Experiment on creating many tokens
1. Timing performance on local vs on td-meta1
    1.  network?
    1. server load?

hectorcorrea:
I've marked my PR as ready for review (https://github.com/pulibrary/tiger-data-app/pull/129) I didn't realize it was still in draft mode.

PR #133, opened 2023-04-10
Adding additional mediaflux/http classes

carolyncole:
Added Session timeout check
Removed class level service_args concept
Also enabled WebMock everywhere instead of enabling it and disabling it 
Additionally changed the test Mediaflux location not to be 0.0.0.0 so WebMock would mock it

refs #116

### DIFF
--- a/config/mediaflux.yml
+++ b/config/mediaflux.yml
@@ -26,7 +26,7 @@ development:
 test:
   api_root_ns: '/td-test-001'
   api_transport: 'http'
-  api_host: '0.0.0.0'
+  api_host: 'mediaflux.example.com'
   api_port: '8888'
   api_domain: 'system'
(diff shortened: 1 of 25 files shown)


config/mediaflux.yml
```diff
@@ -26,7 +26,7 @@ development:
 test:
   api_root_ns: '/td-test-001'
   api_transport: 'http'
-  api_host: '0.0.0.0'
+  api_host: 'test.mediaflux.com'
```
escowles, line 29: Would it make sense to use an `example.com` domain to make sure it never tries to contact a real MediaFlux server? `test.mediaflux.com` doesn't resolve now, but they could register it if they wanted.

hectorcorrea (approved):

Issue #128, opened 2023-03-27
Consolidate documentation for running MediaFlux in Docker locally


kelynch:
Create a document or section of the README that consolidates and offers step-by-step instructions to get a Docker container up and running locally.  Include directions for where to go in the browser to visit the MediaFlux web UI and the aterm interface.

## Acceptance Criteria

- [x] Document where to get local deployment resources
- [x] Document how to deploy and interact with the container
- [x] Document how to get to concepts guide, MFlux documentation, other relevant actions
- [x] Point to internal notes from workshops for exercises

## References: 

* https://pulibrary.slack.com/archives/C02PHG9MGU8/p1679672664170159 
* https://github.com/pulibrary/tiger-data-app#mediaflux-server 
* https://github.com/pulibrary/tiger-data-app/issues/26

PR #135, opened 2023-04-11
First pass at getting up and running for local development

kelynch:
Work toward #128

### DIFF
--- a/docs/local_development.md
+++ b/docs/local_development.md
@@ -52,3 +52,10 @@ This README describes the recommended approach to get up and running with MediaF
    ```
 1. At the login screen, change "server" to `0.0.0.0` and "port" to `8888`.  Use the domain, user, and password information in LastPass to log in.
 
+## Internal documentation
+
+All notes below are internal and cannot be accessed outside of the PUL IT group.
+
+* [Training notes](https://drive.google.com/drive/folders/1kG6oJBnGqOUdM2cHKPxCOC9fBmAJ7iDo)
+* [Hector's training notes](https://drive.google.com/drive/folders/1HGPp43OcGikdZmr3Wd4tgdpY6m1y_PCx)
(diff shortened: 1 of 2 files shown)


hectorcorrea (approved):

docs/local_development.md
```diff
@@ -0,0 +1,61 @@
...
+     WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
+     root@123abc:/setup#
+   ```
+   You are now inside of the Docker container running MediaFlux.
+1. Inside the container, run the following command:
+   ```bash 
+     /usr/bin/env java -jar /usr/local/mediaflux/bin/aserver.jar application.home=/usr/local/mediaflux nogui
+   ```
+   This will start the MediaFlux service.  Leave the service running in this tab.
+
```
carolyncole, line 29: I find it simpler to not run the docker in interactive mode.  It will then just start up mediaflux.
```suggestion
    ```bash 
      docker run --publish 0.0.0.0:8888:8888 $DOCKER_IMAGE_REGISTRY/$NAMESPACE
    ```
    Where `$DOCKER_IMAGE_REGISTRY` and `$NAMESPACE` are replaced with values found in LastPass in the previous step.
1. Upon running the Docker command, you should see output like the following:
   ```bash 
     WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
     root@123abc:/setup#
   ```
    This will start the MediaFlux service.  Leave the service running in this tab.

carolyncole (approved):
@kelynch it is up to you, but I put in a suggestion for just running the docker instance instead of utilizing the interactive commands.  You can keep the interactive commands if you like though...

hectorcorrea:
It might not be a bad idea to add the commands that @carolyncole suggested but I thought we could do that as a separate PR.

Issue #132, opened 2023-04-10
Remove MediafluxClient from Projects and other concept classes


carolyncole:
Instead of utilizing the default MediafluxClient, utilize a session object passed in from the controller
Then change the concept classes to call the Mediaflux::Http classes instead of the mediaflux-client
Finally remove the MediafluxClient class

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/132?planning-poker) with Zenhub @bess @hectorcorrea @jrgriffiniii

PR #159, opened 2023-08-29
Adding mediaflux session information into the session cookie

carolyncole:
This keeps the login around between pages requests

Also removed the mediaflux client from the Welcome controller

refs #132

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,8 +2,20 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   before_action :authenticate_user!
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -2,8 +2,20 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   before_action :authenticate_user!
+  around_action :mediflux_session
```
hectorcorrea, line 5: Was this meant to be `mediaflux_session` ?

hectorcorrea (approved):

Issue #142, opened 2023-04-19
User Mediaflux Login Experiments


carolyncole:
Experiment with Generating Tokens, or creating a login via mediaflux for the user ([Like the catalog does with alma](https://catalog.princeton.edu/users/sign_in?origin=%2Fredirect-to-alma)

Experiment with the login options of mediaflux and how we can tie them to our Princeton users in the tigerdata application.

Acceptance Criteria:

- [x] create a /docs document that lists the options we tried
- [x] Make a comment on  https://github.com/pulibrary/tiger-data-app/issues/153 with findings
- [x] Do performance testing on each option in the document.  One login vs 1K logins via each method
       - [ ] Do Performance Testing from local mediaflux to local application
       - [ ] Do the performance testing over the network testing on tdmeta1 and staging
- [x] Document how long each one is valid (session is valid for 30 minutes without any intervention)

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/142?planning-poker) with Zenhub @bess @hectorcorrea @jrgriffiniii

PR #168, opened 2023-08-31
Adding login options documentation

carolyncole:
Also adds rake task to start the timing process

We do not have access to create tokens for other users on tdmeta1, so experiments on tdmeta1 will have to wait until we get access to `secure.identity.token.create` as an administrator

refs #142


app/models/mediaflux/http/create_token_request.rb
```diff
@@ -0,0 +1,37 @@
...
+      end
+
+      def initialize(session_token:, domain:, user:)
+        @domain = domain
+        @user = user
+        super(session_token: session_token)
+      end
+
+      def identity
+        @id ||= response_xml.xpath("/response/reply/result/token").text.strip
```
jrgriffiniii, line 18: 👍

app/models/mediaflux/http/logon_request.rb
```diff
@@ -50,9 +58,13 @@ def response_session_token
         def build_http_request_body(name:)
           super do |xml|
             xml.args do
-              xml.domain self.class.mediaflux_domain
-              xml.user self.class.mediaflux_user
-              xml.password self.class.mediaflux_password
+              if @identity_token.nil?
```
jrgriffiniii, line 61: 👍

jrgriffiniii (approved):

Issue #169, opened 2023-09-01
TigerData schema - projects


hectorcorrea:
Define via Ruby a schema in MediaFlux for TigerData projects. The following Aterm script shows how the fields could be defined:

```
# Defines a metadata namespace with one document type for "project" data
asset.doc.namespace.create :namespace tigerdata_schema :description "TigerData metadata schema"

asset.doc.type.update :create true :description "Project metadata" :type tigerdata_schema:project :definition < \
    :element -name id            -type string -index true -min-occurs 1 -max-occurs 1 -label "The unique identifier for the project" \
    :element -name title         -type string             -min-occurs 1 -max-occurs 1 -label "A plain-language title for the project" \
    :element -name description   -type string             -min-occurs 1 -max-occurs 1 -label "A brief description of the project" \
    :element -name data_sponsor  -type string -index true -min-occurs 1 -max-occurs 1 -label "The person who takes primary responsibility for the project" \
    :element -name data_manager  -type string -index true -min-occurs 1 -max-occurs 1 -label "The person who manages the day-to-day activities for the project" \
    :element -name data_users_rw -type string -index true -min-occurs 0               -label "A person who has read and write access privileges to the project" \
    :element -name data_users_ro -type string -index true -min-occurs 0               -label "A person who has read-only access privileges to the project" \
    :element -name departments   -type string -index true -min-occurs 1               -label "The primary Princeton University department(s) affiliated with the project" \
    :element -name created_on    -type date               -min-occurs 1 -max-occurs 1 -label "Timestamp project was created" \
    :element -name created_by    -type string             -min-occurs 1 -max-occurs 1 -label "User that created the project" >
```

The code should only create the schema (`asset.doc.namespace.create`) if it does not exist. 

The official schema document is here: https://docs.google.com/document/d/1XjXkzhs5ngM5cmyrmTLaIxG6TF4jT7-HMsMrjj3akRw/edit#heading=h.jyc9l9aomuz

PR #178, opened 2023-09-07
Moved schema definition to its own class

hectorcorrea:
Moved the schema code to its own class and added tests for the schema namespace and schema field definition classes.

This is a follow up to https://github.com/pulibrary/tiger-data-app/pull/175 

Closes https://github.com/pulibrary/tiger-data-app/issues/169

### DIFF
--- a/app/models/tigerdata_schema.rb
+++ b/app/models/tigerdata_schema.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+# Defines the Tigerdata schema in MediaFlux along with the
+# document types allowed. At this point we only support "projects"
+class TigerdataSchema
+  attr_accessor :schema_name, :schema_description
+
+  def initialize(schema_name: nil, schema_description: nil, session_id:)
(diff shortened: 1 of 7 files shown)


app/models/tigerdata_schema.rb
```diff
@@ -0,0 +1,53 @@
...
+      document: "project",
+      description: "Project Metadata",
+      fields: project_schema_fields,
+      session_token: @session_id
+    )
+    fields_request.resolve
+  end
+
+  def project_schema_fields
+    # TODO: I think these fields should be their own Ruby class and we can make them
```
carolyncole, line 38: @hectorcorrea is this todo still relevant?

carolyncole (approved):
There may be an extraneous todo in there, but this looks like a nice step forward

app/models/tigerdata_schema.rb
hectorcorrea: I think so. I'll leave it for now and if by the time we are done messing up with the schema we don't need it we can remove it.

PR #182, opened 2023-09-08
Sample script to create the schema on td-meta1

hectorcorrea:
@carolyncole This is the script that Robert could run on `td-meta` with his credentials to create the `tigerdata` schema.


carolyncole (approved):
Thank you @hectorcorrea

PR #187, opened 2023-09-13
Adding the concept of service execute with a token to the login experiments

carolyncole:
Adds the Service Execute request to our arsenal of mediaflux commands


app/models/mediaflux/http/service_execute_request.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+module Mediaflux
+  module Http
+    class ServiceExecuteRequest < Request
```
jrgriffiniii, line 4: 💯

jrgriffiniii (approved):

PR #190, opened 2023-09-21
Project edit

hectorcorrea:


app/models/project.rb
```diff
@@ -1,97 +1,22 @@
 # frozen_string_literal: true
-class Project
```
hectorcorrea, line 2: The `Project` class is now an ApplicationRecord class and the previous project functionality that interacts with Mediaflux has been moved to `ProjectMediaflux`

bess (approved):
🚀

Issue #173, opened 2023-09-01
User dashboard


hectorcorrea:
Update the home page to display a list of projects the current user is sponsoring as shown in the screenshot on https://github.com/pulibrary/tiger-data-app/issues/113.

The list of projects a user sponsors is captured in the Project record (see https://github.com/pulibrary/tiger-data-app/issues/169). We'll need to come up with the query to fetch records based on the field in the schema.

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-Workcycles-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/173?planning-poker) with Zenhub @bess @jrgriffiniii @kelynch

hectorcorrea:
The code to fetch records for the current user was added on https://github.com/pulibrary/tiger-data-app/pull/190

PR #191, opened 2023-09-27
Moving Mediaflux version information to it's own page

carolyncole:
Remove this information from the dashboard as it is not really part of it

refs #173


hectorcorrea (approved):
I love this. You read my mind, I've been thinking that moving the MF info from the home page will also make the system run even if MF is acting up, and this PR also takes care of that.

PR #198, opened 2023-09-28
Makes sure directory field is passed even when not editable

hectorcorrea:
Fixes a bug in the edit process. Added a test to catch it in the future if we reintroduce it.


app/views/projects/_edit_form.html.erb
```diff
@@ -21,7 +21,12 @@
   </div>
 
   <% if @project.in_mediaflux? %>
-    <p>Project Directory: /tigerdata/<%= @project.metadata[:directory] %> (MediaFlux id: <%= @project.mediaflux_id %>)<br/>
+    <!--
```
jrgriffiniii, line 24: I am very sorry for missing this.

jrgriffiniii (approved):

PR #201, opened 2023-09-29
Removing the multi dashboard concept

carolyncole:
Each user may have many roles so their dashboard will likely reflect all those roles 
The current multi dashboard concept with a different dashboard per role does not match up with the wireframes. 
Instead of maintaining code we do not need I am removing it.


hectorcorrea (approved):
Should we also delete the route that points to this controller? https://github.com/pulibrary/tiger-data-app/blob/main/config/routes.rb#L19

Issue #224, opened 2023-10-18
Performance test for collection queries


kelynch:
From [Feature Refinement notes](https://docs.google.com/document/d/14SjoKlTI7AWNVNjgBOa28jDnhClTnaMnEgN5EBocR9o/edit#heading=h.f1qpx09g69fj):

Run an experiment on a group of collections as follows:

Run both of these queries separately: 
* Query by data sponsor and collect time
* Query by department and collect time

On the same group of collection in the following three settings:
* All collection under 1 namespace
* All collection under a distributed namespace
* All collections under their own namespace

hectorcorrea:
Creating 1000 projects in staging under the same namespace (e.g. `td-demo-001/project1...1000`) works rather well. Creating the projects took less than 2 minutes and queries come back within a few milliseconds

```
# 1,000 projects in staging

Creating 1000 projects with prefix oct26b...
Creating projects 16397.64 ms 163.98 seconds (~2 minutes)

Getting counts by data_sponsor zz001 16.13 ms 0.16 seconds
Getting counts by data_sponsor zz003 35.74 ms 0.36 seconds
Getting counts by data_sponsor zz007 66.79 ms 0.67 seconds
Getting counts by data_sponsor  10.60 ms 0.11 seconds

{:data_sponsor=>"zz001", :count=>"578"}
{:data_sponsor=>"zz003", :count=>"289"}
{:data_sponsor=>"zz007", :count=>"143"}
{:data_sponsor=>"total", :count=>"1015"}

Getting counts by data_sponsor zz007 department THREE 17.24 ms 0.17 seconds
47 records for data_sponsor zz007 department THREE
```

*Creating 10,000 projects under the same namespace on the other hand showed some problems*. It took about 30 minutes to create them (which is no big deal since we won't be creating 10,000 projects at once often) but the larger issue was that queries then started to take 1 to 2 seconds (rather than milliseconds).

```
# 10,000 projects in staging
Creating 10000 projects with prefix oct26c...
Creating projects 186119.23 ms 1861.19 seconds (~30 minutes)

Getting counts by data_sponsor zz001 235.55 ms 2.36 seconds
Getting counts by data_sponsor zz003 235.05 ms 2.35 seconds
Getting counts by data_sponsor zz007 231.55 ms 2.32 seconds
Getting counts by data_sponsor  118.83 ms 1.19 seconds

{:data_sponsor=>"zz001", :count=>"6293"}
[...]

PR #225, opened 2023-10-19
Creating lots of projects to benchmark performance

hectorcorrea:
Adding rake tasks to create a bunch of projects (namespaces + collections) in MediaFlux and validate how well it works on each case.

The rake task can be run as:

```
bundle exec rake projects:create_many[100,test-a]
```

By default it will create the projects under the default root namespace `td-demo-001`. If you want to use a different default root you can specify it via the ENV variable:

```
MEDIAFLUX_ROOT_NS=/td-demo-06 bundle exec rake projects:create_many[100,test-a]
```

You can also run queries against MediaFlux after the projects have been created to see the timings with different parameters:

```
bundle exec rake projects:query[zz007,THREE]
```

See https://github.com/pulibrary/tiger-data-app/issues/224#issuecomment-1781792431 for the results in stating.

### DIFF
--- a/app/models/mediaflux/http/collection_count_request.rb
+++ b/app/models/mediaflux/http/collection_count_request.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+module Mediaflux
+  module Http
+    class CollectionCountRequest < Request
+
+      # Constructor
+      # @param session_token [String] the API token for the authenticated session
+      # @param namespace [String] Namespace to limit the search
+      # @param data_sponsor [String] Datasponsor for the collection
+      # @param data_sponsor [String] Department assigned to the collection
+      def initialize(session_token:, namespace:, data_sponsor: nil, department: nil)
+        super(session_token: session_token)
+        @namespace = namespace
+        @data_sponsor = data_sponsor
+        @department = department
+      end
+
+      # Specifies the Mediaflux service to use
+      # @return [String]
+      def self.service
+        "asset.count"
+      end
+
+      def count
+        @count ||= response_xml.xpath("/response/reply/result/total").text
+        @count
+      end
+
+      private
+
+        def build_http_request_body(name:)
+          super(name: name) do |xml|
+            xml.args do
+              xml.namespace @namespace
+              xml.where build_where_clause
+            end
+          end
+        end
+
+        def build_where_clause
+          where_clause = "asset is collection"
+          if @data_sponsor
+            where_clause += " AND tigerdata:project/data_sponsor='#{@data_sponsor}'"
+          end
+          if @department
+            where_clause += " AND tigerdata:project/departments='#{@department}'"
+          end
+          where_clause
+        end
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


app/models/organization.rb
```diff
@@ -71,12 +71,5 @@ def self.create_root_ns(session_id:)
 
   def self.create_defaults(session_id:)
     create_root_ns(session_id: session_id)
-    organizations = []
-    organizations << { name: "rc", title: "Research Computing" }
-    organizations << { name: "pppl", title: "Princeton Physics Plasma Lab" }
-    organizations << { name: "pul", title: "Princeton University Library" }
-    organizations.each do |org|
-      Organization.create!(org[:name], org[:title], session_id: session_id)
-    end
```
hectorcorrea, line 80: We are not using these anymore so I am preventing the code from creating them. I created a separate ticket to do a more in depth review of this class: https://github.com/pulibrary/tiger-data-app/issues/228

spec/fixtures/files/asset_update_request.xml
```diff
@@ -5,13 +5,13 @@
...
-          <tigerdata:data_manager>data_manager</tigerdata:data_manager>
-          <tigerdata:departments>RDSS</tigerdata:departments>
-          <tigerdata:departments>HPC</tigerdata:departments>
+          <code>code</code>
+          <title>title</title>
+          <description>description</description>
+          <data_sponsor>data_sponsor</data_sponsor>
+          <data_manager>data_manager</data_manager>
+          <departments>RDSS</departments>
+          <departments>HPC</departments>
```
hectorcorrea, line 14: This is very important. Our original code was nesting the metadata properties and that was preventing MediaFlux queries from picking them up.

app/models/mediaflux/http/collection_count_request.rb
```diff
@@ -0,0 +1,51 @@
...
+        super(session_token: session_token)
+        @namespace = namespace
+        @data_sponsor = data_sponsor
+        @department = department
+      end
+
+      # Specifies the Mediaflux service to use
+      # @return [String]
+      def self.service
+        "asset.count"
```
hectorcorrea, line 20: Just gets the count, not the actual assets.

lib/tasks/projects.rake
```diff
@@ -0,0 +1,107 @@
+# frozen_string_literal: true
+namespace :projects do
+  desc "Times the creation of projects and querying by TigerData metadata fields"
+  task :create_many, [:count, :prefix] => [:environment] do |_, args|
+    count = (args[:count] || "").to_i
```
carolyncole, line 5: I'm sure the code you have works, but it would read better to put `raise "Count must be specified" if count.blank?` and then you would not have to deal with the empty string parsing
```suggestion
    raise "Count must be specified" if count.blank?
    count = args[:count].to_i
```

lib/tasks/projects.rake
```diff
@@ -0,0 +1,107 @@
...
+    project_prefix = args[:prefix]
+    raise "Count must be specified" if count == 0
```
carolyncole, line 7: If you go with the above suggestion this line would be removed
```suggestion
```

carolyncole (approved):
@hectorcorrea nice work.  One small change of you want it...

PR #231, opened 2023-10-31
Save the read only and read write users to MF

### LINKED ISSUE #222: Save read_only and read_write users to MediaFlux
Now that we are allowing the user to enter a list of data readers and data writers to the project (#171) we should also save that data to MediaFlux.

The tigerdata/project schema has two fields already defined for this: `data_users_rw` and `data_users_ro` with cardinality 0-N.

### LINKED ISSUE #228: Re-evaluate Organizations
The code currently has a concept of Organizations that we initially used to group projects under. We are not using that concept anymore but the code still has the code https://github.com/pulibrary/tiger-data-app/blob/main/app/models/organization.rb

We should revisit this code and remove it / refactor it to new classes if not needed.

hectorcorrea:
Saves the project read only and read write users to Mediaflux. We were already allowing the user to enter this data, we just were not saving it to MF. Closes #222 

I also removed the obsolete Organization model class (closes #228) and kept in the `Project` model the only part of that code that we really needed which is the code to make sure the root namespace is created before a project is created.


app/models/project_mediaflux.rb
```diff
@@ -50,4 +55,18 @@ def self.safe_name(name)
...
+    if namespace_request.exists?
+      Rails.logger.info "Root namespace #{root_namespace} already exists"
+    else
+      Rails.logger.info "Created root namespace #{root_namespace}"
+      namespace_request = Mediaflux::Http::NamespaceCreateRequest.new(namespace: root_namespace, description: "TigerData client app root namespace",
+                                                                      store: Store.all(session_id: session_id).first.name,
+                                                                      session_token: session_id)
+      namespace_request.response_body
+    end
+  end
```
hectorcorrea, line 71: This is the only useful part from the now defunct Organization model.

bess (approved):

PR #234, opened 2023-11-02
Removed the .yarn folder and .yarn configuration files to have yarn install in npm_modules

### LINKED ISSUE #232: Make yarn install into node_modules
Currently node installs into a .yarn directory.  In all our other projects it installs into node_modules.

We should make this application match our standard.

carolyncole:
Also add npm_modules to the shared directory

fixes #232 

Makes the installation consistent with other projects.


config/deploy.rb
```diff
@@ -37,4 +39,6 @@
     end
   end
 end
+
+before 'deploy:reverted', 'npm:install'
```
carolyncole, line 43: Added because of https://github.com/capistrano/npm/issues/7#issuecomment-274749621

hectorcorrea (approved):

PR #236, opened 2023-11-06
Started adding code to add read-only and read-write users via JavaScript

hectorcorrea:


app/views/layouts/application.html.erb
```diff
@@ -7,9 +7,13 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
 
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap-icons@1.7.1/font/bootstrap-icons.css">
+
     <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
     <%= vite_client_tag %>
     <%= vite_javascript_tag 'application' %>
+
+    <script src="https://code.jquery.com/jquery-3.6.0.min.js"></script>
```
carolyncole, line 16: I surprised you need this. we have jquery included here:https://github.com/pulibrary/tiger-data-app/blob/724245ed452bf2e80e113d573fa8ca055ee2cd00/vite.config.ts#L8

hectorcorrea: I wonder if it is because we kept the JavaScript on the page instead of passing it through the asset pipeline. I've added a ticket to move the JavaScript out (https://github.com/pulibrary/tiger-data-app/issues/296). 

I'll make a note to check if we can remove the extra jQuery line once we implement that ticket.

carolyncole (changes requested):
I see there is a plan to move the JS to it's own file.  I would prefer for it to live with the javascript of the application until it gets moved to it's own file, so that it can be linted.  I'm also unsure why you need to reload jquery, since it is already loaded...  I would approve the PR without the second loading of jquery, since there is already a ticket for the javascript update.

app/views/layouts/application.html.erb
hectorcorrea: Fixed with @carolyncole PR

Issue #245, opened 2023-11-27
Complete aterm 101 tutorial


bess:
Complete the aterm 101 tutorial if you're new, or if it's been awhile: https://github.com/pulibrary/tiger-data-app/blob/main/docs/aterm_101.md

- [x] Bess
- [x] Robert
- [x] Jaymee
- [ ] Kate
- [x] Carolyn
- [x] James
- [ ] Add your name here if you want to

PR #300, opened 2023-12-11
Reflect the fact that aterm.jar is part of the project now

bess:
Ref #245


hectorcorrea:
Oops...I am not sure we should have the aterm.jar on the project! I will suggest not having it on Github.

Issue #253, opened 2023-11-29
Validation for the "Project Directory" field on the new project request form


matthewjchandler:
### User Story
As a Data Sponsor, when I am filling out the "Project Directory" field on the new project request form, I need it to stop me immediately if I enter a value that could cause problems on some or all systems, so I don't have to go back and forth with the system administrators after submitting the form.

### Notes
Following up from #170, the project directory field needs to go a step further to validate entries to be file-path safe. We can adjust this in the future, but for now, my recommendation is to enforce Unicode, limit the entry to 64 characters total, and prohibit characters other than alphanumeric (case-sensitive), minus-dash, and underscore. If you have a suitable alternative to writing new code to do this explicitly, that's fine. We just need a way to nip problematic characters in the bud.

![image](https://github.com/pulibrary/tiger-data-app/assets/66128387/2546899b-e90d-44a6-9e55-f56b06fc0f53)

### Completion Criteria
- [x] The "Project Directory" field is required and singular
- [x] The "Project Directory" field only accepts a string of up to 64 Unicode alphanumeric, minus-dash, and/or underscore characters (or a suitable alternative method for ensuring that values are file path-safe on all systems)
- [ ] Current system of surfacing issues is fine, just augment it to add the character limitations
- [ ] Testing must account for the fact that some directories with be capitalized

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/253?planning-poker) with Zenhub @bess @hectorcorrea @JaymeeH @jrgriffiniii @kelynch @leefaisonr

PR #316, opened 2023-12-14
Introducing directory path validation for project forms

jrgriffiniii:
Advances #253

### DIFF
--- a/app/views/projects/_edit_form.html.erb
+++ b/app/views/projects/_edit_form.html.erb
@@ -19,6 +19,14 @@
     margin-top: -15px;
     margin-bottom: 10px;
   }
+
+  input:valid {
+    background-color: palegreen;
+  }
+
+  input:invalid {
+    background-color: lightpink;
+  }
 </style>
 
 <div>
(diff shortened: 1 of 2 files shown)


app/views/projects/_edit_form.html.erb
```diff
@@ -82,7 +90,8 @@
       (MediaFlux id: <%= @project.mediaflux_id %>)<br/>
       This project has already been saved to Mediaflux and the directory cannot be changed</p>
   <% else %>
-    <p>Project Directory: <%= Rails.configuration.mediaflux["api_root_ns"] %>/<input type="text" id="directory" name="directory" required value="<%= @project.metadata[:directory] %>" /></p>
+    <!-- Unicode alphanumeric, minus-dash, and/or underscore characters -->
+    <p>Project Directory: <%= Rails.configuration.mediaflux["api_root_ns"] %>/<input type="text" id="directory" name="directory" required value="<%= @project.metadata[:directory] %>" pattern="[\w\p{L}\-]{1,64}" /></p>
```
carolyncole, line 94: ❤️

spec/system/project_spec.rb
```diff
@@ -180,18 +181,39 @@
...
         click_on "New Project"
         expect(page.find("#data_sponsor").value).to eq sponsor_user.uid
+        fill_in "data_sponsor", with: sponsor_user.uid
         fill_in "data_manager", with: "xxx"
         fill_in "ro-user-uid-to-add", with: read_only.uid
         click_on "btn-add-ro-user"
         fill_in "rw-user-uid-to-add", with: read_write.uid
         click_on "btn-add-rw-user"
-        fill_in "directory", with: "test_project"
+        fill_in "directory", with: "test-project"
```
carolyncole, line 190: I'm unsure why this change here?  Wouldn't this still be the passing test?

spec/system/project_spec.rb
```diff
@@ -180,18 +181,39 @@
...
+        sign_in sponsor_user
+        visit "/"
+        click_on "New Project"
+        fill_in "data_sponsor", with: sponsor_user.uid
+        fill_in "data_manager", with: data_manager.uid
+        fill_in "ro-user-uid-to-add", with: read_only.uid
+        click_on "btn-add-ro-user"
+        fill_in "rw-user-uid-to-add", with: read_write.uid
+        click_on "btn-add-rw-user"
+        fill_in "directory", with: "test?project"
```
carolyncole, line 210: Could you check the validity value here? @bess put in an example here https://github.com/pulibrary/tiger-data-app/blob/main/spec/system/project_spec.rb#L175

carolyncole (changes requested):
I think this looks good, but I would expect the passing test to still be passing and would like to see you test the validation explicitly

carolyncole (approved):

PR #320, opened 2023-12-15
changing button type for better ui

JaymeeH:
small change for ui

### DIFF
--- a/app/views/projects/_edit_form.html.erb
+++ b/app/views/projects/_edit_form.html.erb
@@ -99,5 +99,5 @@
 
 <div class="actions">
   <input type="submit" class="btn btn-primary" value="Submit" />
-  <%= link_to "Cancel", root_path, class: "btn btn-primary" %>
+  <%= link_to "Cancel", root_path, class: "btn btn-second" %>
 </div>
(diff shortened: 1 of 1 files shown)


carolyncole:
Would be good if the button had a border so it is visible as a button:

Screenshot 2023-12-15 at 1 39 02 PM

app/views/projects/_edit_form.html.erb
```diff
@@ -99,5 +99,5 @@
 
 <div class="actions">
   <input type="submit" class="btn btn-primary" value="Submit" />
-  <%= link_to "Cancel", root_path, class: "btn btn-primary" %>
+  <%= link_to "Cancel", root_path, class: "btn btn-second" %>
```
carolyncole, line 102: I think you meant...
```suggestion
  <%= link_to "Cancel", root_path, class: "btn btn-secondary" %>
```

carolyncole (changes requested):

app/views/projects/_edit_form.html.erb
JaymeeH: your right, pushing a fix for the typo

carolyncole (approved):

Issue #257, opened 2023-11-29
Preliminary process to assign a draft DOI to validated new project requests


matthewjchandler:
### User Story
As a Data Sponsor, when I submit a form for a new project request, I need the system to assign a persistent identifier to the project automatically, so that I don't have to try to figure out some bespoke way to keep track of my project as it evolves within the research data ecosystem.

### Notes
We want to establish an automated process to reserve draft DOIs from DataCite for new TigerData projects. This should only happen after the preliminary new project request form validates all of the entered metadata (i.e., it depends on #256), because we do not want to proliferate draft DOIs unnecessarily.

While a DOI is still in draft form, DataCite does not require much metadata, and we will allow most project metadata to be revised, so I recommend we do not push much to DataCite upon reserving the draft. I suggest that providing the Data Sponsor's name as the first Creator for DataCite will be sufficient for now (but I could be convinced to even leave that out, because technically, we are only collecting NetIDs, and we may not want to do a name lookup before reserving a draft DOI).

The reserved DOI has to be stored in the project metadata in the "Project ID" field, using the standard DOI format: [prefix]/[suffix]. "Project ID" is a required field for every project, even though it is not shown in the new project request form.

API keys added in separate ticket: https://github.com/pulibrary/tiger-data-app/issues/325

### Completion Criteria
- [ ] After validating the entered values from the preliminary new project request form, a draft DOI gets reserved from DataCite for a submitted new project
- [ ] It makes a draft DOI after a completely valid form has been submitted
- [ ] The reserved draft DOI gets recorded in the "Project ID" field for the submitted new project

bess:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/257?planning-poker) with Zenhub @carolyncole @hectorcorrea @JaymeeH @leefaisonr

carolyncole:
I'm reopening this issue because I put the DOI creation after project approval, not project creation.  It needs to be moved...

PR #331, opened 2024-01-05
Copying PULDatacite service class from PDC Describe

carolyncole:
Adjusting the classes to not publish the DOI since there is no current requirement for that functionality

refs #257


hectorcorrea:
Nice!

hectorcorrea (approved):

Issue #261, opened 2023-11-29
Clicking the "Return to Dashboard" button returns the user to the preliminary dashboard page, where now the title of the newly created project, with an indicator for "Pending", is appended to the list of projects in the "My Sponsored Projects" section


matthewjchandler:

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/261?planning-poker) with Zenhub @hectorcorrea @JaymeeH @jrgriffiniii @kelynch @leefaisonr

PR #332, opened 2024-01-05
adding methods to project model and updating the project spec to chec…

JaymeeH:
…k for additional roles
ref #261

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -35,6 +35,15 @@ class Project < ApplicationRecord
     Project.where("metadata_json->>'data_sponsor' = ?", sponsor)
   end
 
+  def self.managed_projects(manager)
+    Project.where("metadata_json->>'data_manager' = ?", manager)
+  end
+
+  def self.data_user_projects(user)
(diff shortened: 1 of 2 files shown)


app/models/project.rb
```diff
@@ -35,6 +35,18 @@ def self.sponsored_projects(sponsor)
...
+  def self.managed_projects(manager)
+    Project.where("metadata_json->>'data_manager' = ?", manager)
+  end
+
+  def self.data_user_projects(user)
+    # See https://scalegrid.io/blog/using-jsonb-in-postgresql-how-to-effectively-store-index-json-data-in-postgresql/
+    # for information on the @> operator
+    query_ro = '{"data_user_read_only":["' + user + '"]}'
+    query_rw = '{"data_user_read_write":["' + user + '"]}'
+    Project.where("(metadata_json @> ? :: jsonb) OR (metadata_json @> ? :: jsonb)", query_ro, query_rw)
```
hectorcorrea, line 47: We are no sure this is the best SQL to query both fields, but it does the job.

bess: First get it working, then later get it working fast if performance is a problem. :) 
LGTM.

bess: Also, hooray for the inline comment pointing to documentation.

bess (approved):
🎉

Issue #287, opened 2023-12-06
Preliminary project metadata model


matthewjchandler:
### User Story
As a Data Sponsor, I need the metadata I entered into the new project request form to be formatted and supplemented with defaults and/or system-generated values automatically, and then sent to the system administrators automatically, so that I can easily submit my requests and not have to worry about unfamiliar technical details.

### Notes
Following up on #170 and #195, I want to be explicit about the expectations for project metadata records. The other issues related to #113 lay out specific requirements for each metadata field in this preliminary phase, according to the [TigerData Standard Metadata Schema v0.5](https://docs.google.com/document/d/1XjXkzhs5ngM5cmyrmTLaIxG6TF4jT7-HMsMrjj3akRw/edit#heading=h.acxfulp3a5og). Right now, the only required early adopter field that is not a part of the form is "Project ID", which is system-generated, following #257. As we move forward, however, there will be more required fields, along with some optional fields, that have default and/or system-generated values. The standard metadata schema will also evolve over time. So we need to be careful to establish an adaptable, extensible framework modeling project metadata.

I don't have requirements for how exactly your database is set up to be adaptable and extensible. For me right now, the deliverable that will help me see that we are modeling metadata in accordance with the roadmap from the Metadata Working Group is a JSON or XML file produced for each submitted new project request.

(This issue supersedes and replaces #258.)

### Completion Criteria
~~- [ ] Upon submission and validation, the values entered in the new project request form are combined with the system-generated Project ID (DOI) to compose a complete preliminary project metadata record~~ https://github.com/pulibrary/tiger-data-app/issues/336
- [x] The preliminary project metadata record can be formatted as XML for Mediaflux commands
- [x] The preliminary project metadata record can be formatted as JSON schema in addition (human readable)
- [x] .xml on the record show page shows the Mediaflux XML
- [x] .json on the record show page shows the JSON
- [x] The xml file should be available as an attachment to the email
- [x] The json file should be available as an attachment to the email
- [x] During the preliminary testing phase, a XML version of the metadata record for each requested new project is emailed to the PO "Create this new project. see attached XML". The email has two attachments XML & JSON

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/287?planning-poker) with Zenhub @JaymeeH @jrgriffiniii @hectorcorrea

[...]

PR #337, opened 2024-01-09
Adding json and xml formats response for project show

carolyncole:
refs #287

WIP

### DIFF
--- a/app/models/project_mediaflux.rb
+++ b/app/models/project_mediaflux.rb
@@ -44,6 +44,16 @@ class ProjectMediaflux
     values
   end
 
+  def self.xml_payload(project:, xml_namespace: nil)
+    project_name = safe_name(project.directory)
+    project_namespace = "#{Rails.configuration.mediaflux['api_root_ns']}/#{project_name}-ns"
+
+    tigerdata_values = project_values(project: project)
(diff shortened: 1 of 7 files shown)


bess (commented):
Looks great!

hectorcorrea:
Nice, I like the XML payload idea to allow us to export what we send to Mediaflux.

hectorcorrea (approved):

bess (approved):

app/models/project_mediaflux.rb
```diff
@@ -43,6 +44,16 @@ def self.project_values(project:)
     values
   end
 
+  def self.xml_payload(project:, xml_namespace: nil)
```
hectorcorrea, line 47: @carolyncole In looking at this, I think this is returning the XML to create the namespace only, but the original code (https://github.com/pulibrary/tiger-data-app/blob/main/app/models/project_mediaflux.rb#L3-L21) sends two requests to Mediaflux: One to create the namespace and one to create the collection assets underneath. I am not sure how we want to represent this in the what will produce for people to reuse.

carolyncole: @hectorcorrea correct.  It only creates the collection asset.create XML, not the namespace create XML.

carolyncole: We could possible attach 3 files to the email, one for the collection, one for the namespace, and then the json representation of the data.

PR #344, opened 2024-01-11
Added XML and json attachments to the create email 

### LINKED ISSUE #287: Preliminary project metadata model

carolyncole:
Also added to_xml and to_json to the project to allow reuse of the code
Also added commented out mailcatcher configuration for development
Also renamed the text version of the email

fixes #287

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -21,11 +21,10 @@ class ProjectsController < ApplicationController
     respond_to do |format|
       format.html
       format.json do
-        render json: @project.metadata
+        render json: @project.to_json
       end
       format.xml do
-        xml_payload = ProjectMediaflux.xml_payload(project: @project)
-        render xml: xml_payload
+        render xml: @project.to_xml
       end
     end
   end
--- a/app/mailers/tigerdata_mailer.rb
+++ b/app/mailers/tigerdata_mailer.rb
@@ -3,6 +3,23 @@ class TigerdataMailer < ApplicationMailer
   def project_creation
     config = Rails.application.config.tigerdata_mail[:project_creation]
     @project = params[:project]
+
+    filebase = @project.metadata[:project_id].tr("/", "_")
+
+    # attaching json response  to the mailer
+    json_content = @project.to_json
+    attachments["#{filebase}.json"] = {
+      mime_type: "application/json",
+      content: json_content
+    }
+
+    # attaching xml response to the mailer
+    xml_content = @project.to_xml
+    attachments["#{filebase}.xml"] = {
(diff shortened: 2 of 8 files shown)


app/controllers/projects_controller.rb
```diff
@@ -21,11 +21,10 @@ def show
     respond_to do |format|
       format.html
       format.json do
-        render json: @project.metadata
+        render json: @project.to_json
       end
       format.xml do
-        xml_payload = ProjectMediaflux.xml_payload(project: @project)
-        render xml: xml_payload
+        render xml: @project.to_xml
```
bess, line 27: 👍🏻 I like this pattern.

bess (approved):
⭐

PR #350, opened 2024-01-16
Ensuring that Project departments are sorted and comma-separated for the show view

jrgriffiniii:
Resolves #269

PR #356, opened 2024-01-16
Extending the project form input validation

jrgriffiniii:
Ensuring that the project form validation does not raise errors for cases where the data sponsors and data managers are specified, but the read-only and read/write users are not. These changes were necessary to propose when testing the changes found within https://github.com/pulibrary/tiger-data-app/pull/350


bess (approved):
Great testing. Thank you.

Issue #358, opened 2024-01-17
Update the sender of the email


bess:
The email that we sent is currently from "example.com". It should come from no-reply@princeton.edu

bess:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/358?planning-poker) with Zenhub @carolyncole @hectorcorrea @JaymeeH @leefaisonr @kelynch

jrgriffiniii:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/358?planning-poker) with Zenhub @jrgriffiniii

PR #365, opened 2024-01-18
update sender of  email

Seanwarren77:
ref #358


carolyncole (changes requested):
I would really love to see a test to verify this change.  I would suggest checking the from in https://github.com/pulibrary/tiger-data-app/blob/main/spec/mailers/tigerdata_spec.rb#L13

Something like
```
expect(mail.from).to eq ["no-reply@princeton.edu"]
```

carolyncole (approved):
Thank you!

PR #380, opened 2024-01-23
correcting the time zone of the provenance creation date

### LINKED ISSUE #378: Provenance time zone should be set to Princeton time
timezone

JaymeeH:
closes #378


carolyncole (approved):
I assume there is already a test that includes the date format?  Since Matt was particular it would be good if there was one.

PR #406, opened 2024-01-25
Integrating Sidekiq as the default ActiveJob processor

jrgriffiniii:
Advances #388


carolyncole (changes requested):
@jrgriffiniii can you add the redis configuration to the qa environment too? https://github.com/pulibrary/tiger-data-app/blob/main/config/environments/qa.rb#L62

carolyncole (approved):

PR #422, opened 2024-02-02
Adding an accumulator task so folks can create accumulators

carolyncole:
An example of running it in the console...
```
user = User.first
collection = 1234 # exisiting collection id
accum = Mediaflux::Http::CreateCollectionAccumulatorRequest.new(session_token: user.mediaflux_session, collection:, name: test-asset-create-count2,  type:collection.asset.count)
accum.resolve
accum.response_xml
```
Should give you something like
```
<response><reply type="result"><result><added id="1197">7</added></result></reply></response>
```

### DIFF
--- a/app/models/mediaflux/http/create_collection_accumulator_request.rb
+++ b/app/models/mediaflux/http/create_collection_accumulator_request.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+module Mediaflux
+  module Http
+    class CreateCollectionAccumulatorRequest < Request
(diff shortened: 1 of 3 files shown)


hectorcorrea (approved):

hectorcorrea:
This is fantastic @carolyncole 🎉

JaymeeH:
Note: If an accumulator is added to a project after assets have already been added/generated -- We will need to regenerate the accumulator to add all previously created files/collections to the count.

The request in the console... 
`asset.collection.accumulator.all.generate :id [_collection_id_]   `

PR #429, opened 2024-02-05
Fix mediaflux query to include all children and sub-collections

carolyncole:
This more closely mimics having specified collection in the top level of the xml

### DIFF
--- a/app/models/mediaflux/http/query_request.rb
+++ b/app/models/mediaflux/http/query_request.rb
@@ -51,7 +51,7 @@ module Mediaflux
               # TODO: there is a bug in mediaflux that does not allow the comented out line to paginate
               #      For the moment we will utilize the where clasue that does allow pagination
               # xml.collection collection if collection.present?
-              xml.where "asset in collection #{collection}" if collection.present?
+              xml.where "asset in collection or subcollection of #{collection}" if collection.present?
               xml.where aql_query if aql_query.present?
               xml.action action if action.present?
               xml.idx idx
--- a/spec/models/mediaflux/http/query_request_spec.rb
+++ b/spec/models/mediaflux/http/query_request_spec.rb
@@ -43,7 +43,7 @@ RSpec.describe Mediaflux::Http::QueryRequest, type: :model do
       before do
         stub_request(:post, mediflux_url)
           .with(body: "<?xml version=\"1.0\"?>\n<request>\n  <service name=\"asset.query\" session=\"secretsecret/2/31\">\n    "\
-                      "<args>\n      <where>asset in collection 1067</where>\n      <idx>1</idx>\n      <size>10</size>\n    </args>\n  </service>\n</request>\n")
+                      "<args>\n      <where>asset in collection or subcollection of 1067</where>\n      <idx>1</idx>\n      <size>10</size>\n    </args>\n  </service>\n</request>\n")
           .to_return(status: 200, body: query_response, headers: {})
       end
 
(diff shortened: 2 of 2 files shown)


app/models/mediaflux/http/query_request.rb
```diff
@@ -51,7 +51,7 @@ def build_http_request_body(name:)
               # TODO: there is a bug in mediaflux that does not allow the comented out line to paginate
               #      For the moment we will utilize the where clasue that does allow pagination
               # xml.collection collection if collection.present?
-              xml.where "asset in collection #{collection}" if collection.present?
+              xml.where "asset in collection or subcollection of #{collection}" if collection.present?
```
hectorcorrea, line 54: I think this was meant to be `asset.query :collection #{collection} if collection.present?` but notice that this would in addition to whatever `where` clause we use.

hectorcorrea (commented):
I think the `:collection` parameter would go in addition to the `:where` parameter

hectorcorrea (changes requested):
I think the :collection parameter would go in addition to the :where parameter

hectorcorrea:
@carolyncole it looks like there is a reason we are not using the `:collection` parameter to query. There is a bug in Mediaflux that prevents pagination from working when we use the `:collection` parameter.

See note here: https://github.com/pulibrary/tiger-data-app/blob/main/app/models/mediaflux/http/query_request.rb#L51-L53

You can reproduce the bug as follows:

```
asset.query :collection 1098 :idx 3 :size 3
 :id -version "1" "1099"
    :id -version "1" "1100"
    :id -version "1" "1101"
    :cursor
        ...

asset.query :collection 1098 :idx 4 :size 3
    :id -version "1" "1099"
    :id -version "1" "1100"
    :id -version "1" "1101"
    :cursor
        ...
```

Notice how Mediaflux returned the exact same asset ids even when we asked for the second page of results.

app/models/mediaflux/http/query_request.rb
carolyncole: @hectorcorrea I'm just adding `or subcollection of` so that the updated code to get around the bug mimics what happens with adding `:collection`. The secondary where clause is still included.

carolyncole:
> @carolyncole it looks like there is a reason we are not using the `:collection` parameter to query. There is a bug in Mediaflux that prevents pagination from working when we use the `:collection` parameter.
> 
> See note here: https://github.com/pulibrary/tiger-data-app/blob/main/app/models/mediaflux/http/query_request.rb#L51-L53
> 
> You can reproduce the bug as follows:
> 
> ```
> asset.query :collection 1098 :idx 3 :size 3
>  :id -version "1" "1099"
>     :id -version "1" "1100"
>     :id -version "1" "1101"
>     :cursor
>         ...
> 
> asset.query :collection 1098 :idx 4 :size 3
>     :id -version "1" "1099"
>     :id -version "1" "1100"
>     :id -version "1" "1101"
>     :cursor
>         ...
> ```
> 
> Notice how Mediaflux returned the exact same asset ids even when we asked for the second page of results.

@hectorcorrea I did not change it to `:collection` because there is a bug.  `:collection` queries both the collection and all sub collections, so the old code was not equivalent to `:collection`. You need `asset in collection or subcollection of` to mimic the responses you getback when `:collection` is included.  This is ONLY an update to make the work around in our code work the same way as the `:collection`. I am unsure what your objection is.  This is truely confusing...  There is still the additional where clause.  I abosolutely did not change that.  Otherwise the PR would show it...

Issue #445, opened 2024-02-07
Roles in the user should only be define by the information in data/user_registration_list.csv


carolyncole:
The config files https://github.com/pulibrary/tiger-data-app/blob/main/config/default_mediaflux_admins.yml & https://github.com/pulibrary/tiger-data-app/blob/main/config/default_sponsors.yml are obsolete

We should remove them from the system and utilize the data from the user table (eligible_sponsor & eligible_manager)

Acceptance criteria
- [x] Anyone in the spreadsheet has a user account
- [x] The code allows anyone with a User model to login to the system
- [x] The user eligible_sponsor & eligible_manager columns are populated from the spreadsheet
- [x] Remove the other config files and rake tasks associated

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/445?planning-poker) with Zenhub @bess @hectorcorrea @JaymeeH @jrgriffiniii @leefaisonr

PR #450, opened 2024-02-08
Removing roles so the eligible columns control who can login

### LINKED ISSUE #415: Add role columns from User Registration List to the User model
Completion Criteria: 

- [x] User model has fields for `eligible_sponsor` and `eligible_manager` 
- [x] User model is populated when user registration csv is loaded 
- [ ] Factory for user has the ability to create `eligible_sponsor` and `eligible_manager` users

carolyncole:
fixes #415

refs #445

Also removes the code that creates a system user for every person who logs in with a cas account.  This will not allow us to rely on folks in the database being eligible data users


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -1,17 +1,18 @@
...
+    access_token = request.env["omniauth.auth"]
+    @user = User.from_cas(access_token)
 
-    if @user.nil?
+    if @user.nil? && access_token&.provider == "cas"
+      redirect_to help_path
+      flash.notice = "You can not be signed in at this time."
+    elsif @user.nil?
       redirect_to root_path
       flash.alert = "You are not a recognized CAS user."
```
hectorcorrea, line 12: ah...this is the error that Chuck got yesterday during the meeting.

hectorcorrea (approved):

hectorcorrea:
This is great...I think we should also remove or rewrite the "Structure" section of the readme (
https://github.com/pulibrary/tiger-data-app?tab=readme-ov-file#structure) but can do that as a separate PR.

Issue #393, opened 2024-01-24
SPIKE: how to query mediaflux for storage usage?


bess:
![Screenshot 2024-01-24 at 11 34 46 AM](https://github.com/pulibrary/tiger-data-app/assets/65608/64371314-8fe5-4f5b-a8c1-ad381cc611ae)

- This is extremely similar to #392, and the goal is to ensure that this is actually possible, and to determine how difficult this is to implement
- This might involve Accumulators, talk to Chuck
- Scope one week for addressing this
- Query to see what the current quota set is, and then query for what is actually being used for storage
- Physical use space vs. logical use space
  - Do we need a convention on how to report disk/storage usage?
  - Working through this should uncover aspects to this question

bess:
It appears that if storage has been allocated to a collection, we will get back a `quota` stanza from aterm:
![quota_in_aterm](https://github.com/pulibrary/tiger-data-app/assets/65608/fa6b5855-bf5e-4ccf-9d3c-bead5c8d7794)

PR #458, opened 2024-02-08
Build mediaflux request to add a quota to a collection

bess:
Ref #393


spec/fixtures/files/create_collection_quota_response.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<response>
+    <reply type="result">
+        <result></result>
```
hectorcorrea, line 4: wow...surprised that their response it's empty!

hectorcorrea (approved):

PR #467, opened 2024-02-09
Ensuring that superusers can create projects from the welcome dashboard

jrgriffiniii:
By default, even with the `superuser` role, one still cannot create new projects.


leefaisonr (approved):
Looks great - Thanks!!

PR #468, opened 2024-02-09
Added Chuck

### LINKED ISSUE #439: Add Chuck Bentler to MediaFlux admins
Add Chuck (netid `cbentler`) [here](https://github.com/pulibrary/tiger-data-app/blob/main/config/default_mediaflux_admins.yml#L21).

kelynch:
Closes #439

### DIFF
--- a/config/default_mediaflux_admins.yml
+++ b/config/default_mediaflux_admins.yml
@@ -2,6 +2,7 @@
 production:
   - bs3097 # Bess Sadler
   - cac9 # Carolyn Cole
+  - cbentler # Chuck Bentler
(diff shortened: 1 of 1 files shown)


bess (changes requested):
@kelynch I thought we had decided to remove this file in favor of streamlining all of the auth configuration to one place? See this ticket: https://github.com/pulibrary/tiger-data-app/issues/445

However, even after that ticket was closed I see that the `config/default_mediaflux_admins.yml` file is still present in the code base. Seems like we've got a disconnect about what the expected behavior should be. I'm going to put something on the meeting agenda so we can iron that out.

PR #469, opened 2024-02-09
Downloads file list to CSV

### LINKED ISSUE #398: SPIKE: how do we get all of the files for the project and put them in a CSV?
- [ ] CSV needs three headers: path; size; and last modified
- [ ] All project files must be included
- [ ] OK to use mediaflux default values (i.e., don't worry about converting dates to ISO 8601)
- [ ] Should be constructed with a background job (separate ticket)
- [ ] Determine how long it will take (rough estimate by file count, e.g. 1,000 vs 100,000)
- [ ] Determine what the feasible implementations are in Mediaflux, and see if one shines as the best
- [ ] Output could be a write-up on the options (with outline of the processes); if multiple, spell out pros and cons

hectorcorrea:
Added rake task to download the *entire file list* from a Project to a CSV file:

```
rake projects:download_file_list[netid,project-id]
```

where `netid` is your netid and `project-id` is the PostgreSQL id of the project in the database. File list is saved to file `file_list_[project-id].csv`

The down download process uses `iterators` and fetches 1000 files at a time. It takes about 10 minutes to download 215,000 files from `projectg` in `td-meta1` to my local machine. It produces a 23MB text file.

I also modified the code to always fetch the file list using iterators and ditched our previous code that used cursors since iterators seem to be the way to go for what we are intending to do (cursors are a better option if we want to navigate back and forth through the file list, iterators are better for read forward only)

This download process will eventually need to be moved to a background job since it takes too long but that is a separate ticket.

Closes #398


app/models/mediaflux/http/query_request.rb
```diff
@@ -29,28 +25,17 @@ def self.service
...
-
-        # total is only the actual total when the "complete" attribute is true,
-        # otherwise it reflects the total fetched so far
-        {
-          ids: ids,
-          size: xml.xpath("/response/reply/result/size").text.to_i,
-          cursor: parse_cursor(xml.xpath("/response/reply/result/cursor")),
-          files: parse_files(xml)
-        }
+        xml.xpath("/response/reply/result/iterator").text.to_i
```
hectorcorrea, line 31: This is the "big" change. It returns an iterator (rather than a cursor). We then use the iterator to fetch the data via calls to `IteratorRequest` as shown in `app/models/project.rb` below.

app/models/project.rb
```diff
@@ -93,14 +93,56 @@ def asset_count(session_id:)
...
-      deep_search: true)
-    request.result
+  # Fetches the first n files
+  def file_list(session_id:, size: 10)
+    return { files: [] } if mediaflux_id.nil?
+
+    query_req = Mediaflux::Http::QueryRequest.new(session_token: session_id, collection: mediaflux_id, deep_search: true)
+    iterator_id = query_req.result
+
+    iterator_req = Mediaflux::Http::IteratorRequest.new(session_token: session_id, iterator: iterator_id, size: size)
```
hectorcorrea, line 103: Fetch the data using the iterator.

app/models/project.rb
```diff
@@ -93,14 +93,56 @@ def asset_count(session_id:)
...
+    return { files: [] } if mediaflux_id.nil?
+
+    query_req = Mediaflux::Http::QueryRequest.new(session_token: session_id, collection: mediaflux_id, deep_search: true)
+    iterator_id = query_req.result
+
+    File.open(filename, "w") do |file|
+      # file header
+      file.write("ID, PATH, NAME, COLLECTION?, LAST_MODIFIED, SIZE\r\n")
+      loop do
+        iterator_req = Mediaflux::Http::IteratorRequest.new(session_token: session_id, iterator: iterator_id, size: 1000)
```
hectorcorrea, line 125: A more complete example, fetches the data from the iterator until the iterator says it has fetched all the data (completed)

PR #470, opened 2024-02-12
updating header logo

JaymeeH:
photoshop the tigerdata logo to be viewable on the new black header on tigerdata.


carolyncole (approved):
Thanks for grabbing the new logo from the tigerdata site.

Issue #402, opened 2024-01-25
Enforce roles and permissions


bess:
## Acceptance criteria
- [x] Add use cases for all of the following to `spec/system/role_enforcment_spec.rb` (or whatever this file is called) 
- [x] Ensure that only people eligible to be a `Data Sponsor` can request a project
- [x] The `Data Sponsor` field is not editable. It is automatically assigned and cannot be changed. 
- [x] Ensure that only the `Data Sponsor` of a specific project can:
  - [x] add a `Data Manager`
  - [x] remove a `Data Manager`
  - [x] change a `Data Manager`
- [x] Ensure that only people eligible to be a `Data Manager` can be added in the `Data Manager` role
- [x] Ensure that only people with a `Data Sponsor` or `Data Manager` role can add or edit a `Data User`

bess:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/402?planning-poker) with Zenhub @jrgriffiniii @hectorcorrea @leefaisonr @JaymeeH

PR #480, opened 2024-02-14
Allow superusers to make a new project

bess:

Ref #402


carolyncole (approved):
Good catch!

PR #492, opened 2024-02-15
Only eligible Data Managers can be added in the Data Manager role

leefaisonr:
Refs #402

### DIFF
--- a/spec/factories/user.rb
+++ b/spec/factories/user.rb
@@ -21,12 +21,6 @@ FactoryBot.define do
 
     ##
     # A user who is allowed to manage a project
-    factory :project_manager do
-      eligible_manager { true }
-    end
-
-    ##
-    # A user who is allowed to manage a project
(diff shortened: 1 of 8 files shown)


spec/factories/user.rb
```diff
@@ -25,6 +25,12 @@
       eligible_manager { true }
     end
 
+    ##
+    # A user who is allowed to manage a project
+    factory :data_manager do
```
carolyncole, line 30: We should probably just change project_manager to be named data_manager so we do not confuse things by having two names for the same concept.

carolyncole (changes requested):
Looking great!  Just one small tweak...

spec/factories/user.rb
leefaisonr: Thank you @carolyncole ! - I have made those requested changes.

carolyncole (approved):

PR #505, opened 2024-02-19
Details page - Edit buttons

### LINKED ISSUE #273: Project Details page - edit buttons
### User Story
As a Data Manager, when I review a project's current metadata and settings on the Project Details page, I need a clear indication of which sections I am permitted to edit, so that I can make any necessary changes quickly.

### Notes
Data Managers have some limited privileges to edit project attributes, in order to allow them to manage the day-to-day aspects of an ongoing project. These privileges include adding and removing Data Users (not changing the Data Sponsor or Data Manager), as well as editing the Description fields.

Data Sponsors also have full permissions to edit project attributes. Their user journey will be addressed later, but for now, the "Edit" buttons described in this issue apply to both Data Sponsors and Data Managers.

When the viewer of the Project Details page is the Data Manager or Data Sponsor for that project, they should see clickable "Edit" buttons next to the "Roles" and "Description" headings, like so:

TD_ProjectDetails_EditButtons_2024-01-23

Clicking an Edit button should take one to the "Project Revision Request" form, as detailed in #275.

### Completion Criteria

- [ ] If a project is still pending, no "Edit" buttons are shown
- [ ] If the project is active and the user is a Data Sponsor or Data Manager for the project, they can see "Edit" buttons next to the "Roles" and "Description" headings on the "Project Details" page
- [ ] Clicking either "Edit" button takes the user to the "Project Revision Request" (separate issue)

JaymeeH:
added edit buttons next to the project role and description headings
edit buttons link to the edit form
edit buttons are only accessible to sponsors and managers 
edit buttons are only displayed on active projects

closes #273

### DIFF
--- a/app/views/projects/show.html.erb
+++ b/app/views/projects/show.html.erb
@@ -12,7 +12,7 @@ Project Details:
 
 <div class="details">
   <h2>Roles
-    <% if @project.metadata[:status] == @approve_status && @eligible_editor %>
+    <% if @project_eligible_to_edit %>
       <%= link_to "Edit", edit_project_path(@project), id:"revisible-heading", class: "btn btn-primary btn-sm" %>
     <% end %>
   </h2>
@@ -45,7 +45,7 @@ Project Details:
 
 <div class="details">
   <h2>Project Description
-    <% if @project.metadata[:status] == @approve_status && @eligible_editor %>
+    <% if @project_eligible_to_edit %>
       <%= link_to "Edit", edit_project_path(@project), id:"revisible-heading", class: "btn btn-primary btn-sm" %>
     <% end %>
   </h2>
@@ -93,11 +93,6 @@ Project Details:
 
 <div class="details">
   <%= link_to "Review Contents", project_contents_path(@project), class: "btn btn-primary btn-sm" %>
-  
-  <% if @project.metadata[:status] == @approve_status %>
-    <%= link_to "Edit Project", edit_project_path(@project), class: "btn btn-primary btn-sm" %>
-  <% end %>
-
   <%= link_to "Return to Dashboard", root_path, class: "btn btn-primary btn-sm" %>
   <%= link_to "Withdraw Project Request", "", class: "btn btn-light btn-sm" %>
 </div>
--- a/app/views/projects/show.html.erb
+++ b/app/views/projects/show.html.erb
@@ -11,7 +11,11 @@ Project Details:
 <% end %>
 
 <div class="details">
-  <h2>Roles</h2>
+  <h2>Roles
+    <% if @project.metadata[:status] == @approve_status && @eligible_editor %>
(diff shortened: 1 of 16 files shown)


app/views/projects/show.html.erb
```diff
@@ -40,7 +44,11 @@ Project Details:
 </div>
 
 <div class="details">
-  <h2>Project Description</h2>
+  <h2>Project Description
+    <% if @project.metadata[:status] == @approve_status && @eligible_editor %>
```
carolyncole, line 48: @JaymeeH I'm not going to hold up the PR, but another time you could have put this logic in the controller.  Logic in the view is a lot harder to test.  I might have added `@project_eligble_to_edit` flag in the controller and then used it in the two places in the view.

spec/system/project_show_spec.rb
```diff
@@ -31,13 +31,16 @@
     context "Navigation Buttons" do 
       it "Shows the correct nav buttons for an approved project" do
         sign_in sponsor_user
+        sponsor_user["eligible_sponsor"] = true
```
carolyncole, line 34: The factory for a sponsor user (on line 6) should be utilized in this test instead of setting the flag. 
```
let(:sponsor_user) { FactoryBot.create(:project_sponsor, uid: "pul123") }
```

spec/system/project_show_spec.rb
```diff
@@ -31,13 +31,16 @@
...
+        sponsor_user.save!
         project_in_mediaflux.metadata_json["status"] = Project::APPROVE_STATUS
         project_in_mediaflux.save!
         visit "/projects/#{project_in_mediaflux.id}"
         
         expect(page).to have_content(project_in_mediaflux.title)
         expect(page).not_to have_content(pending_text)
-        expect(page).to have_link("Edit")
+        expect(page).to have_link("Edit") #button next to role and description heading
+        expect(page).to have_link("Edit Project")
```
carolyncole, line 43: Is this the original Edit button?  I believe that should be removed.

carolyncole (changes requested):
@JaymeeH this is great work.  Just a couple of small comments on the code, which I would not block the PR for.  But I do believe the original edit button should be removed.  Do with what you like on the stylistic comments.

app/views/projects/show.html.erb
bess: +1 to put this in the controller. There are going to be lots of places where we will need to test whether the project is eligible to edit, and the criteria for that will likely evolve over time. Let's put the logic in a single place where it can be tested.

spec/system/project_show_spec.rb
bess: I agree with Carolyn. Again, the characteristics of a project sponsor are likely to change over time, so using the factory will keep our tests more robust as things inevitably shift.

bess (changes requested):
Looks great overall, but I agree with Carolyn on a few changes needed. Thank you!

bess (approved):

Issue #410, opened 2024-01-26
Project schema mismatch


hectorcorrea:
We made the Project description optional in the Work edit form but the Mediaflux schema still has this field marked as required. Notice the `min-occurs` is still 1 in https://github.com/pulibrary/tiger-data-app/blob/main/app/models/tigerdata_schema.rb

```
description = { name: "description", type: "string", index: false, "min-occurs" => 1, "max-occurs" => 1, label: "A brief description of the project" }
```

bess:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/410?planning-poker) with Zenhub @hectorcorrea @JaymeeH @jrgriffiniii @leefaisonr @carolyncole

PR #507, opened 2024-02-19
Tigerdata Schema: make description an optional field

JaymeeH:
updated the tigerdata schema to make the description optional
updated the spec that generates the aterm command to create the schema
closes: #410


carolyncole (approved):

JaymeeH:
Note: We did not ping Chuck/Robert to run the script since they are now defining the schema slightly different from what we need

PR #509, opened 2024-02-20
Create a test to create a project in local mediaflux

carolyncole:
This test without the mocks
refs #410

### DIFF
--- a/docs/schema_script.txt
+++ b/docs/schema_script.txt
@@ -1,4 +1,4 @@
-# This file was automatically generated on 2024-01-18T08:45:16-05:00
+# This file was automatically generated on 2024-02-16T16:36:12-05:00
 # Create the "tigerdata" namespace schema and the "project" definition inside of it.
 #
 # To run this script, issue the following command from Aterm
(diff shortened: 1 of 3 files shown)


docs/schema_script.txt
```diff
@@ -12,7 +12,7 @@ asset.doc.namespace.update :create true :namespace tigerdata :description "Tiger
 asset.doc.type.update :create true :description "Project metadata" :type tigerdata:project :definition < \
  :element -name code -type string -index true -min-occurs 1 -max-occurs 1 -label "The unique identifier for the project" \
  :element -name title -type string -min-occurs 1 -max-occurs 1 -label "A plain-language title for the project" \
- :element -name description -type string -min-occurs 1 -max-occurs 1 -label "A brief description of the project" \
+ :element -name description -type string -min-occurs 0 -max-occurs 1 -label "A brief description of the project" \
```
bess, line 15: I'm confused. I thought you said it IS required. Doesn't this change it to not being required? "min-occurs 0" means "not required", right?

spec/system/project_spec.rb
```diff
@@ -268,6 +268,50 @@
...
+        @original_api_host = Rails.configuration.mediaflux["api_host"]
+        Rails.configuration.mediaflux["api_host"] = "0.0.0.0"
+        @session_id = sponsor_user.mediaflux_session
+      end
+
+      after do
+        Rails.configuration.mediaflux["api_host"] = @original_api_host
+      end
+
+      it "allows the projects to be created" do
```
bess, line 283: Same question... if description is required, then when the description is empty it should NOT allow the project to be required, right?

bess (changes requested):
Less of a "request changes" and more of an "I'm confused." I thought we were making description required? But this seems to do the opposite? What am I missing?

bess (approved):
Nevermind. I got mixed up between "Description" and "Department"

PR #511, opened 2024-02-20
Implements feature show more/less files 

### LINKED ISSUE #390: Add a "show more" button to the project contents page
Add a “show more” button that displays up to 100 files

Screenshot 2024-01-24 at 11 30 33 AM

- Please note that DataTables does *not* have a tree view for the assets
- We may want more columns within the DataTables; There will need to be a research ticket for exploring additional metadata for the columns
  - This may include replacing DataTables
- aterm command: `asset.query :action get-name` or `asset.query :action get-meta`

Acceptance Criteria:
- [ ] This retrieves the second set of files from the Mediaflux server API
- [ ] If there are not more than 100 files, then the "Show More" button is in the disabled HTML state (greyed out)
- [ ] DataTables is updated with the second set of files retrieved from Mediaflux

hectorcorrea:
Allows the user to view more (or less) files from the Mediaflux backend for the current project.

## Initial display
Screenshot 2024-02-21 at 10 18 23 AM

## After Show more
Screenshot 2024-02-21 at 10 18 31 AM

Closes #390

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -85,7 +85,7 @@ class ProjectsController < ApplicationController
       project.metadata_json["status"] = Project::APPROVE_STATUS
       params.delete("mediaflux_id")
     end
-    
+
     #Edit action
     if params.key?("title")
       project_metadata = ProjectMetadata.new(project: project, current_user:)
@@ -95,7 +95,7 @@ class ProjectsController < ApplicationController
       })
       project.metadata = project_metadata.update_metadata(params: metadata_params)
     end
-    
+
     # @todo ProjectMetadata should be refactored to implement ProjectMetadata.valid?(updated_metadata)
     if project.save
       redirect_to project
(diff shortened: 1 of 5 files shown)


app/models/mediaflux/asset.rb
```diff
@@ -12,6 +12,17 @@ def initialize(id:, name:, collection:, path: nil, last_modified_mf: nil, size:
       @last_modified_mf = last_modified_mf
     end
 
+    def name
```
hectorcorrea, line 15: This change was technically not part of the original issue, but our TestGenerator uses Mediaflux's unnamed files so I had to update the code to support them. Which in the end is a good thing since now our code handles these files in case our users were to use this feature in Mediaflux.

app/controllers/projects_controller.rb
```diff
@@ -118,7 +118,7 @@ def revision_confirmation; end
   def contents
     project
     @num_files = project.asset_count(session_id: current_user.mediaflux_session)
-    @file_list = project.file_list(session_id: current_user.mediaflux_session, size: 20)
+    @file_list = project.file_list(session_id: current_user.mediaflux_session, size: 100)
```
hectorcorrea, line 121: Fetching 20 or 100 files takes just the same amount of time since the bulk of it is creating the HTTP request to Mediaflux. Therefore we fetch 100 files to begin with and show only the first 10 to the user and show the other 90 via JavaScript once they click the "Show More" button.

bess (approved):

PR #512, opened 2024-02-20
Handle Redis client errors within ProjectsController#create

### LINKED ISSUE #477: [Honeybadger] Improve error handling for Project Creation when Redis Failures are encountered
Please see the following example: https://app.honeybadger.io/projects/113559/faults/104746997#notice-trace

In cases such as these, even if the Redis outage is not related to the Ruby on Rails codebase architecture, there might still be a need to improve error handling for end-users. For example:

```ruby
def create
    new_project
    project_metadata = ProjectMetadata.new( current_user:, project: new_project)
    new_project_params = params.dup
    metadata_params = new_project_params.merge({
      status: Project::PENDING_STATUS
    })
    project_metadata.create(params: metadata_params)
    if new_project.save
      TigerdataMailer.with(project: @project).project_creation.deliver_later
      redirect_to project_confirmation_path(@project)
    else
      render :new
    end
  rescue RedisClient::CannotConnectError => redis_client_error
    error_message = "Failed to create the new project: Failed to connect to the Redis client: #{redis_client_error}"
    Rails.logger.error(error_message)
    flash[:notice] = error_message
    render :new
  end
```

for https://github.com/pulibrary/tiger-data-app/blob/main/app/controllers/projects_controller.rb#L14

Acceptance Criteria:

- [x] Ensure that errors of the type `RedisClient::CannotConnectError` are properly handled from within `ProjectsController#create`
- [x] This error must be logged
- [x] The user should be aware that an error occurred when trying to create the new project

jrgriffiniii:
Ensuring that ProjectsController#create actions which encounter Redis client connection errors log these for Rails, flash a warning for end-users, and render the New Project View. This resolves #477

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -21,12 +21,18 @@ class ProjectsController < ApplicationController
     project_metadata.create(params: metadata_params)
     if new_project.save
       mailer = TigerdataMailer.with(project_id: @project.id)
-      mailer.project_creation.deliver_later
+      message_delivery = mailer.project_creation
+      message_delivery.deliver_later
 
       redirect_to project_confirmation_path(@project)
     else
       render :new
     end
+  rescue RedisClient::CannotConnectError => redis_connect_error
+    error_message = "Failed to connect to Redis: #{redis_connect_error}"
+    Rails.logger.error(error_message)
+    flash[:notice] = error_message
+    render :new
   end
 
   def show
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -29,10 +29,10 @@ class ProjectsController < ApplicationController
       render :new
     end
   rescue RedisClient::CannotConnectError => redis_connect_error
-    error_message = "Failed to connect to Redis: #{redis_connect_error}"
+    error_message = "We are sorry, while the project was successfully created, an error was encountered which prevents the delivery of an e-mail message confirming this. Please know that this error has been logged, and shall be reviewed by members of RDSS."
 
     Rails.logger.error(error_message)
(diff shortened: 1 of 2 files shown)


app/controllers/projects_controller.rb
```diff
@@ -21,12 +21,18 @@ def create
...
+      message_delivery = mailer.project_creation
+      message_delivery.deliver_later
 
       redirect_to project_confirmation_path(@project)
     else
       render :new
     end
+  rescue RedisClient::CannotConnectError => redis_connect_error
+    error_message = "Failed to connect to Redis: #{redis_connect_error}"
+    Rails.logger.error(error_message)
```
bess, line 33: @jrgriffiniii I like the idea of rescuing and logging these errors, but shouldn't we also notify honeybadger? I'm concerned that we never actually look at our rails error log unless we're trying to track down something specific, whereas we should be looking at honeybadger every day to find out if there are problems that need attention. If we had a serious problem with redis we'd likely not know without the honeybadger alerts. 

What happens to the `deliver_later` in this scenario? Does this message get silently dropped? That seems like a problem. Can we talk about what the user experience should be instead? Showing the user a redis connection error seems less than ideal: the user can't do anything with this information. Maybe instead something like "We experienced an error and could not deliver this email message?"

bess (dismissed):
Can we talk as a team about what the behavior should be here? I am going to add it to the standup agenda.

jrgriffiniii:
Following the discussion held during the RDSS Sprint Planning meeting held on 02/21/24, this should be closed in order to proceed with a solution which ensures that end-users do not fail to receive e-mail messages when new projects are created.

carolyncole (approved):

Issue #498, opened 2024-02-16
Standardize time zone


bess:
Currently we're using EST (Eastern Standard Time) (e.g.: `Time.zone.parse(mf_date).in_time_zone("EST").strftime("%Y-%m-%dT%H:%M:%S%:z")`) Seems like this might cause problems during Eastern Daylight Time -- our time stamps will be off by an hour. 

However, just using `ET` (Eastern Time) doesn't work. 

Maybe we want to use `American/New York` instead, like this? https://github.com/pulibrary/tiger-data-app/blob/c93e7b24376c78ec99dc3d0883fee032e7424b03/app/views/projects/show.html.erb#L79

### Acceptance criteria
- [x] Decide on an approach 
- [x] Document the decision in an ADR
- [ ] Implement it the same way everywhere we're using timestamps

kelynch:
Decided to use "American/New York" as the approach.

PR #518, opened 2024-02-21
Adjust time zone to reflect app standardization

leefaisonr:
Ref #498

### DIFF
--- a/app/models/project_metadata.rb
+++ b/app/models/project_metadata.rb
@@ -50,13 +50,13 @@ class ProjectMetadata
         timestamps = {}
         if project.metadata[:created_by].nil?
           timestamps[:created_by] = current_user.uid
-          timestamps[:created_on] = DateTime.now.strftime("%d-%b-%Y %H:%M:%S")
+          timestamps[:created_on] = Time.current.in_time_zone("America/New_York").iso8601
 
         else
           timestamps[:created_by] = project.metadata[:created_by]
           timestamps[:created_on] = project.metadata[:created_on]
           timestamps[:updated_by] = current_user.uid
-          timestamps[:updated_on] = DateTime.now.strftime("%d-%b-%Y %H:%M:%S")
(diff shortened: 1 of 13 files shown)


app/models/project_metadata.rb
```diff
@@ -50,13 +50,13 @@ def project_timestamps
...
           timestamps[:created_by] = current_user.uid
-          timestamps[:created_on] = DateTime.now.strftime("%d-%b-%Y %H:%M:%S")
+          timestamps[:created_on] = Time.current.in_time_zone("America/New_York").iso8601
 
         else
           timestamps[:created_by] = project.metadata[:created_by]
           timestamps[:created_on] = project.metadata[:created_on]
           timestamps[:updated_by] = current_user.uid
-          timestamps[:updated_on] = DateTime.now.strftime("%d-%b-%Y %H:%M:%S")
+          timestamps[:updated_on] = Time.current.in_time_zone("America/New_York").iso8601
```
JaymeeH, line 59: Note: When writing to mediaflux, I think dates have to be in a specific format which is different from this. If it in an unaccepted format— instead of converting to the appropriate format — it will override the created/updated on fields to nil.

spec/factories/project.rb
```diff
@@ -8,8 +8,8 @@
       data_user_read_only { [] }
       data_user_read_write { [] }
       title { FFaker::Movie.title }
-      created_on { DateTime.now }
-      updated_on { DateTime.now }
+      created_on { Time.current.in_time_zone("America/New_York").iso8601 }
+      updated_on { Time.current.in_time_zone("America/New_York").iso8601 }
```
JaymeeH, line 12: we could try creating a project in rails, then manually creating it in medialfux with `ProjectMediaflux.create!` to see if the new date fields persist. Again, not an issue now — but if the plan is to be able to write to mediaflux in the future — then it may be later on.

leefaisonr: Thank you, @JaymeeH we are in the process of fixing this.

PR #519, opened 2024-02-21
Generates file list as background job

hectorcorrea:


hectorcorrea:
Replaced with https://github.com/pulibrary/tiger-data-app/pull/525

PR #530, opened 2024-02-26
Save export files to a shared location

### LINKED ISSUE #523: File list download
Make sure the files [produced when downloading the file list from a project](https://github.com/pulibrary/tiger-data-app/issues/489) are saved to a file on a shared location. Currently the file is saved on the server where the job was run but there is no guarantee that the request to download the file will be processed by the same server.

Ticket to give access to the servers to the shared location: https://github.com/pulibrary/princeton_ansible/issues/4703

~~When should we clean those files? e.g. after x number of days? after download?~~ Moved this to issue #542

hectorcorrea:
Save export files to our shared location so that can be fetched regardless of what server ran the job originally.

Also added a sample rake task to delete old files from shared location (say after 30 days). We could schedule this task to run on schedule (see issue #542):

To view a list of export files:
```
bundle exec rake exports:report_old
```

To delete old export files:
```
bundle exec rake exports:delete_old
```

Closes #523

### DIFF
--- a/app/jobs/list_project_contents_job.rb
+++ b/app/jobs/list_project_contents_job.rb
@@ -7,7 +7,10 @@ class ListProjectContentsJob < ApplicationJob
     raise "Invalid user id #{user_id} for job #{job_id}" if user.nil?
 
     # Queries Mediaflux for the file list and saves it to a CSV file.
-    project.file_list_to_file(session_id: mediaflux_session, filename: filename_for_export)
+    filename = filename_for_export
+    Rails.logger.info "Exporting file list to #{filename} for project #{project_id}"
+    project.file_list_to_file(session_id: mediaflux_session, filename: filename)
+    Rails.logger.info "Export file generated #{filename} for project #{project_id}"
 
     mark_user_job_as_complete(project: project, user: user)
   end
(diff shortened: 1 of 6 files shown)


app/jobs/list_project_contents_job.rb
```diff
@@ -21,12 +23,10 @@ def mediaflux_session
       logon_request.session_token
     end
 
-    # TODO: This only works with one server. In a multi-server environment
-    # there is no guarantee that the client will be able to fetch the file
-    # since requests could go to any of the many servers.
-    # See https://github.com/pulibrary/tiger-data-app/issues/523
     def filename_for_export
-      "#{Dir.pwd}/public/#{job_id}.csv"
+      raise "Shared location is not configured" if Rails.configuration.mediaflux["shared_files_location"].blank?
```
hectorcorrea, line 27: This is the core of this PR. Files are now saved to a shared location so that they can be fetch by any web server when the user wants to download it.

app/services/project_job_service.rb
```diff
@@ -7,6 +7,9 @@ def initialize(project:)
 
   def list_contents_job(user:)
     job = ListProjectContentsJob.perform_later(user_id: user.id, project_id: @project.id)
+    # Log the job id and the Sidekiq JID in case we need to troubleshoot the job
+    # https://github.com/sidekiq/sidekiq/wiki/Active-Job#job-id
+    Rails.logger.info("Job scheduled, job id: #{job.job_id}, (Sidekiq JID: #{job.provider_job_id || 'nil'})")
```
hectorcorrea, line 12: This could be handy in the future so we can know for sure the state of our jobs in Sidekiq. For example we could query Sidekiq to see the status (enqueued, in process, dead). This will require that we save the `JID` value to the database and use the Sidekiq API to query for its status, which is not part of this PR, but something we could do in the future.

bess (approved):

PR #548, opened 2024-02-29
[Spike] What timezone is mediaflux using

### LINKED ISSUE #539: Spike What timezone is mediaflux using
We need to verify where the timezone comes from for the dates in mediaflux.

We hope they are utilizing unix time, but lets check

Acceptance criteria

- [x] Document where the time zone comes from in Mediaflux in the Time Zone ADR

JaymeeH:
updating timestamps ADR to include context about mediaflux default timezones

closes #539


bess (approved):
Great explanation!

PR #556, opened 2024-03-12
Tigerdata Schema Documentation

JaymeeH:
updated the tigerdata schema so each field has the correct label, description, and instructions in both rails and mediaflux


carolyncole (approved):
I ran the script on my local mediaflux and it ran nicely

PR #566, opened 2024-03-13
Updating Mediaflux Timezone Documentation

JaymeeH:
updating the timezone documentation to include the correct timezone for the live td-meta1 environment


JaymeeH:
Putting a pause on this PR. Upon further review, it looks like The server itself and the assets' time zone are UTC.  It will translate when it presents the results based on the session's time zone. further discussion with the team is needed

carolyncole (approved):
@JaymeeH Thank you!

Issue #550, opened 2024-03-07
Build a docker image out of a mediaflux jar


bess:
Translate our automated ansible build and mediaflux configuration into a docker build for local development

- [x] Our local development mflux needs to match test, staging, and production as much as possible
- [x] Write instructions for how the team can access it

bess:
.jar file requested from Chuck and Robert via slack.

bess:
First draft of a docker image here (private repo): https://github.com/PrincetonUniversityLibrary/mflux_ci

However now I'm getting errors when I try to start the development instance of mediaflux, which had previously been working 🤔 : 

> [1 main],version=4.15.004,server,11-Mar-2024 20:41:59.003:INFO:Started
arc.mf.licence.ExInvalidLicence: The licence at file:/usr/local/mediaflux/config/licence.xml cannot be used: not a licence for this host. Expected '02:42:ac:11:00:02' but found the following ids [02:42:ac:11:00:03]

bess:
This compose.yml file will build a mediaflux container that can serve content to a rails container. 
```
services:
  mediaflux:
    mac_address: "02:42:ac:11:00:02"
    image: "princeton_dev:23.08.25"
    ports:
      - "8888:8888"
    networks:
      vlan:
        ipv4_address: 192.168.0.60
  rails:
    image: cimg/ruby:3.2-browsers
    ports:
      - "3000:3000"
    command: tail -f /dev/null # keep the container running
    networks:
      vlan:
        ipv4_address: 192.168.0.61
  # redis:
  #   image: "redis:alpine"


networks:
  vlan:
    driver: macvlan
    driver_opts:
      parent: eth0
    ipam:
      config:
        - subnet: "192.168.0.0/24"
          ip_range: "192.168.0.64/26"
          gateway: "192.168.0.1"
```

bess:
We have part of a process going, here: https://github.com/PrincetonUniversityLibrary/mflux_ci
However, the image that we're building doesn't appear to really work. For example, you can get to the top level webpage at 0.0.0.0:8888, but you cannot get to /desktop or /aterm. 

We are going to ask whether we can get some of Todd's time so he can show us how he made our initial development docker images for mediaflux.

bess:
For now, Todd has given us an updated docker image. Unclear yet whether we'll continue to get these and that's how we'll stay abreast of the version changes in mediaflux production, or whether we'll need to pick this work back up, hopefully with some vendor support.

PR #577, opened 2024-03-18
Run mediaflux in Circle CI on a Machine executor

carolyncole:
This also removes the NO_CI concept as mediaflux tests can be run in CI
refs #550


bess (approved):
🎉

PR #584, opened 2024-03-20
Tigerdata Storage Usage for Projectg in qa

### LINKED ISSUE #581: Projectg accumulator for storage usage is being read incorrectly in QA
Even though projectg has an accumulator for storage usage attached, tigerdata displays its usage as `Storage Usage
0 KB / 500 GB` which is the default value when the accumulator value is "blank". A similar issue is occuring in mediaflux — where the quota attatched to projectg says `0 bytes`

Acceptance Criteria:
- [ ] The Storage Usage component on the project contents page displays the correct Storage Usage on Tigerdata QA
- [ ] The Quota displays the correct Storage Usage in Mediaflux QA

JaymeeH:
Setting the metadata value for the size accumulator when querying a collection in mediaflux

Fixes #581

### DIFF
--- a/spec/models/mediaflux/http/get_metadata_request_spec.rb
+++ b/spec/models/mediaflux/http/get_metadata_request_spec.rb
@@ -24,6 +24,7 @@ RSpec.describe Mediaflux::Http::GetMetadataRequest, type: :model do
       expect(metadata[:collection]).to be_falsey
       expect(metadata[:path]).to eq("/td-test-001/collection-96-55948/file-96-57045")
       expect(metadata[:type]).to eq("content/unknown")
+      expect(metadata[:size]).to be nil
       expect(WebMock).to have_requested(:post, mediflux_url)
     end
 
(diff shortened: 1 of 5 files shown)


spec/models/mediaflux/http/get_metadata_request_spec.rb
```diff
@@ -24,8 +24,6 @@
       expect(metadata[:collection]).to be_falsey
       expect(metadata[:path]).to eq("/td-test-001/collection-96-55948/file-96-57045")
       expect(metadata[:type]).to eq("content/unknown")
-      expect(metadata[:size]).to eq("")
```
carolyncole, line 27: @JaymeeH I would expect the test to make sure the correct value is here...

carolyncole (changes requested):
@JaymeeH nice update.  Can you please update the test to check for a value?

spec/models/mediaflux/http/get_metadata_request_spec.rb
JaymeeH: Gotcha, I'll work on that right now

carolyncole (approved):

JaymeeH:
Source of the Quota bug in Mediaflux:

Anything that is imported as a reference that is not managed by MediaFlux can still be served up through the system but does not count for quota space.  This is kind of a funky thing that I can see causing some confusion in the future but for the moment, all that needs to happen is to copy the projectg data to a managed store.

PR #606, opened 2024-03-25
memoizing mediaflux metadata request

### LINKED ISSUE #604: Project#mediaflux_metadata is not cached
This means every call out to the metadata is a separate mediaflux call.
https://github.com/pulibrary/tiger-data-app/blob/main/app/models/project.rb#L95

JaymeeH:
memoizing the mediaflux metadata method to reduce the number of calls to mediaflux                                                                                                                                                           fixing tiny bug with storage capacity

closes #604

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -92,8 +92,11 @@ class Project < ApplicationRecord
   end
 
   def mediaflux_metadata(session_id:)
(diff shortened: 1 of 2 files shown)


carolyncole (approved):
Great!

PR #641, opened 2024-04-15
refactoring #approve_project

### LINKED ISSUE #639: Refactor the #approve_project method
Look in the projects controller, in the update method.

Screenshot 2024-04-12 at 3 30 23 PM

## Acceptance criteria
- [x] Calling the #approve_project method on a project sets the status to `Project::APPROVED_STATUS`

JaymeeH:
- Calling the #approve_project method on a project sets the status to Project::APPROVED_STATUS
- Calling the #approve_project method on a project records the mediaflux id in `project.mediaflux_id`

closes #639

### DIFF
--- a/spec/models/project_metadata_spec.rb
+++ b/spec/models/project_metadata_spec.rb
@@ -148,7 +148,7 @@ RSpec.describe ProjectMetadata, type: :model do
     it "Records the mediaflux id and sets the status to approved" do 
       project_metadata = described_class.new(current_user: current_user, project:)
       params = {mediaflux_id: 001 }
-      project_metadata.approve_project(params:) # doesn't call the doi service twice
+      project_metadata.approve_project(params:)
       
       project.reload
 


spec/models/project_metadata_spec.rb
```diff
@@ -145,6 +145,16 @@
     end
 
     describe "#approve_project" do
+    it "Records the mediaflux id and sets the status to approved" do 
+      project_metadata = described_class.new(current_user: current_user, project:)
+      params = {mediaflux_id: 001 }
+      project_metadata.approve_project(params:) # doesn't call the doi service twice
```
carolyncole, line 151: @JaymeeH what does the comment mean here?  As far as I can tell the DOI service is not called at all.
```suggestion
      project_metadata.approve_project(params:)
```

carolyncole (changes requested):
@JaymeeH Looks great.  One small change requested

spec/models/project_metadata_spec.rb
JaymeeH: Ah, good catch — looks like that's a comment that was left over from the create method. I can remove it

carolyncole (approved):

JaymeeH:
Thanks for the review @carolyncole!

Issue #615, opened 2024-03-27
Erase all users and projects in QA, both from rails and mediaflux


carolyncole:
Have a rake task that allows us to erase all data on the server that only runs in QA, Dev, staging, and test.  Specifically does not run in production.

Completion Criteria:
- [x] A rake task that errors in production
- [ ] In other environments
     - [ ] cleans the rails database 
     - [ ] the mediaflux server from the root namespace

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/615?planning-poker) with Zenhub @bess @hectorcorrea @JaymeeH @jrgriffiniii @kelynch @leefaisonr

leefaisonr:
This command will delete a namespace AND everything in it on ATerm:

`asset.namespace.hard.destroy :atomic true :force true :namespace /td-test-001/tigerdataNS/Loquat1NS`

PR #653, opened 2024-04-22
Add ability to destroy a namespace and its contents

leefaisonr:
Ref #615


spec/fixtures/files/namespace_destroy_response.xml
bess, line 1: Why is this needed? Looks like the test is actually deleting things from mediaflux and this fixture isn't used?

bess (approved):

PR #654, opened 2024-04-23
Bugfix: Logon resolve method should return session_token

bess:
Ref #615

### DIFF
--- a/app/models/mediaflux/http/logon_request.rb
+++ b/app/models/mediaflux/http/logon_request.rb
@@ -40,8 +40,7 @@ module Mediaflux
         @session_token = nil
         super
         @session_token = response_session_token
-
-        @http_response
+        @session_token
       end
 
       private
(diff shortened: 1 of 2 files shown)


carolyncole (approved):
@bess this PR does not follow the pattern in other models.  Not that the class currently follows the pattern in other models, but we are currently doing something more like adding a methods like https://github.com/pulibrary/tiger-data-app/blob/main/app/models/mediaflux/http/create_asset_request.rb#L32  I'm approving this PR as I'm sure it will work, but just making you aware you are making a change to the overall pattern.  The resolve methods in all other classes returns the response.

bess:
@carolyncole Ah, thank you for telling me that. I will close the PR. I saw a [comment](https://github.com/pulibrary/tiger-data-app/blame/main/app/models/mediaflux/http/logon_request.rb#L38):
`# @return [String] the session token generated by the Mediaflux API`
I thought the method was supposed to return the token but had been broken and no one noticed.

PR #663, opened 2024-04-26
Fixing validation and MetadataError

### LINKED ISSUE #661: Validation is not working
### Expected behavior
I expected to get information on why the call to mediflux was failing

### Actual behavior
Here is the error I got
```
    TigerData::MetadataError:
       Project failed to create with metadata schema version 0.6 with the missing fields: Project directory Missing metadata value for project_directory, Title Missing metadata value for title, Status Missing metadata value for status, Data sponsor Missing metadata value for data_sponsor, Data manager Missing metadata value for data_manager, Department s Missing metadata value for department_s, Created on Missing metadata value for created_on, Created by Missing metadata value for created_by, Storage capacity Missing metadata value for storage_capacity, Storage performance expectations Missing metadata value for storage_performance_expectations, Submission Missing metadata value for submission, Revision s Missing metadata value for revision_s, Schema version Missing metadata value for schema_version
```
The actual error from mediaflux indicated something very different
```
{:title=>"arc.mf.server.Services$ExServiceError", :message=>"call to service 'asset.create' failed: XPath tigerdata:project is invalid: missing element 'Performance'\nContext:\n            26-Apr-2024 12:15:07.477 [0.023 sec(s) ago]: service: system:manager [sid=204]: asset.create :name  \"turkey\" :namespace  \"/td-test-001/tigerdataNS/turkeyNS\" :meta < :tigerdata:project -xmlns:tigerdata \"tigerdata\" -id \"2\" < :ProjectDirectory  \"******\" :Title  \"******\" :Description  \"******\" :Status  \"******\" :DataSponsor  \"******\" :DataManager  \"******\" :Department  \"******\" :Department  \"******\" :CreatedOn  \"******\" :CreatedBy  \"******\" :ProjectID  \"******\" :StorageCapacity < :Size  \"******\" :Unit  \"******\" > :StoragePerformance  \"******\" :ProjectPurpose  \"******\" > > :collection -cascade-contained-asset-index \"true\" -contained-asset-index \"true\" -unique-name-index \"true\"  \"true\" :type  \"application/arc-asset-collection\" :pid  \"path=/td-test-001/tigerdata\" \n  called by: 26-Apr-2024 12:15:07.475 [0.028 sec(s) ago]: http: from: /192.168.65.1, request: /__mflux_svc__ vars:  select headers: [User-Agent=Ruby, Connection=keep-alive, Host=0.0.0.0:8888]\n  called by: 26-Apr-2024 12:15:07.396 [0.107 sec(s) ago]: network: ctime=26-Apr-2024 12:15:07.396, origin=/192.168.65.1, received=3423 [3.42 KB], sent=7991 [7.99 KB]"}
```
### Steps to replicate

In the rails console run the following (this assumes you have created a project in the UI):
```
project = Project.first
project.metadata[:storage_performance_expectations_requested] =nil
session_id = User.first.mediaflux_session
ProjectMediaflux.create!(project:, session_id:)
```

### Impact of this bug
I just put a debugger into the code to get the actual value, so this is impactful, but not very important.

JaymeeH:
fixing metadata error to only pass the missing required field from the project

closes #661

### DIFF
--- a/app/models/project_metadata.rb
+++ b/app/models/project_metadata.rb
@@ -4,11 +4,10 @@ class ProjectMetadata
 
   class Validator < ActiveModel::Validator
     def validate(record)
-      if record.required_attributes.values.include?(nil)
-
+      if record.required_attributes(record).values.include?(nil)
         record.required_keys.each do |attr|
-          value = record.attributes[attr]
-          record.errors.add(attr, "Missing metadata value for #{attr}") if value.nil?
+          value = record.required_attributes(record)[attr]
+          record.errors.add(attr, "Missing metadata value for #{attr}") if value.nil? && record.project.metadata_json.include?(attr)
         end
       end
     end
@@ -76,11 +75,11 @@ class ProjectMetadata
 
   def required_keys
     tableized = required_field_labels.map { |v| v.parameterize.underscore }
-    tableized.map(&:to_sym)
+    tableized
   end
 
-  def required_attributes
-    attributes.select { |k,v| required_keys.include?(k) }
+  def required_attributes(record)
+    record.project.metadata_json.select { |k,v| required_keys.include?(k) }
   end
 
   def attributes
--- a/spec/models/project_mediaflux_spec.rb
+++ b/spec/models/project_mediaflux_spec.rb
@@ -98,7 +98,7 @@ RSpec.describe ProjectMediaflux, type: :model, stub_mediaflux: true do
           collection_id = ProjectMediaflux.create!(project: incomplete_project, session_id: session_token)
         }.to raise_error do |error|
           expect(error).to be_a(TigerData::MetadataError)
-          expect(error.message).to include("Project failed to create with metadata schema version 0.6 with the missing fields:")
+          expect(error.message).to include("Project creation failed with metadata schema version 0.6 due to the missing fields:")
         end
       end
     end
--- a/lib/tiger_data/metadata_error.rb
+++ b/lib/tiger_data/metadata_error.rb
@@ -7,7 +7,7 @@ module TigerData
     def self.missing_metadata(schema_version:, errors:)
       error_messages = errors.full_messages.join
       # include current expected metadata schema version number, as well as what metadata is missing
-      raise TigerData::MetadataError, "Project failed to create with metadata schema version #{schema_version} with the missing fields: #{error_messages}"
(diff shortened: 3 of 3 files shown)


spec/models/project_mediaflux_spec.rb
```diff
@@ -98,7 +98,7 @@
           collection_id = ProjectMediaflux.create!(project: incomplete_project, session_id: session_token)
         }.to raise_error do |error|
           expect(error).to be_a(TigerData::MetadataError)
-          expect(error.message).to include("Project failed to create with metadata schema version 0.6 with the missing fields:")
+          expect(error.message).to include("Project creation failed with metadata schema version 0.6 due to the missing fields:")
```
bess, line 101: Could we add which fields are missing? That seems like an important bit of functionality that we want to ensure we're testing.

bess (changes requested):
Overall this looks great! I asked for one small change, which is that I believe we should also test that the code can return which fields are missing.

spec/models/project_mediaflux_spec.rb
JaymeeH: Yes we can! I'll work on that now

app/models/project_metadata.rb
```diff
@@ -76,11 +75,11 @@ def create( params:)
 
   def required_keys
     tableized = required_field_labels.map { |v| v.parameterize.underscore }
-    tableized.map(&:to_sym)
+    tableized
   end
 
-  def required_attributes
-    attributes.select { |k,v| required_keys.include?(k) }
+  def required_attributes(record)
```
carolyncole, line 81:
```suggestion
  def required_attributes
```

carolyncole (changes requested):
@JaymeeH Looks great!  In addition to @bess request I have a couple suggestions

app/models/project_metadata.rb
```diff
@@ -4,11 +4,10 @@ class ProjectMetadata
 
   class Validator < ActiveModel::Validator
     def validate(record)
-      if record.required_attributes.values.include?(nil)
-
+      if record.required_attributes(record).values.include?(nil)
```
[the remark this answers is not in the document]
JaymeeH: you're absolutely right, i'll fix this

jrgriffiniii (approved):

carolyncole (approved):

bess (approved):

PR #664, opened 2024-04-29
Enable cleanup of a tiger data environment

### LINKED ISSUE #615: Erase all users and projects in QA, both from rails and mediaflux

bess:
Fixes #615


JaymeeH (approved):
Great! Thanks for this!

Issue #629, opened 2024-04-10
Update the system to 0.6 version of the metadata 


carolyncole:
Update to code to produce the metadata to the standard of [Tigerdata Schema](https://docs.google.com/document/d/1RDCowxE2ulfkh0GrqTln8Tr9t5XcWOZ2U--zRpiwe48/edit#heading=h.jyc9l9aomuz)

Completion Criteria:

- [x] Update the Schema create script (including generation code) to include the changes from v0.5 to v0.6
     - [x] focus on early adopter field, not the standard schema
     - [x] Include provenance fields
- [x] Make updates to the internal storage for any changes in the early adopter fields that are necessary
- [x] Run these changes on the mediaflux library sandbox/playground (qa)
- [x] Pass along the script to Chuck incase they need it for td-meta1 (staging)
- [x] No changes to the forms are part of this work, the data from the forms should be massaged into this new version (aka unit for size will be GB and the size will 500)
- [x] Include a version (which is coming in 0.6.1)

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/629?planning-poker) with Zenhub @bess @hectorcorrea @JaymeeH @jrgriffiniii @kelynch @leefaisonr

PR #668, opened 2024-04-30
Updating storage capacity and storage performance to v0.6.1 format

carolyncole:
refs #629

### DIFF
--- a/app/models/project_mediaflux.rb
+++ b/app/models/project_mediaflux.rb
@@ -33,16 +33,18 @@ class ProjectMediaflux
           #All exceptions that are raised should include the current expected metadata schema version number, as well as what metadata is missing.
     id = create_request.id
     if id.blank?
-      response_xml = create_request.response_xml
-      response_text = response_xml.text
-      case response_text
+      response_error = create_request.response_error
+      case response_error[:message]
       when "failed: The namespace #{project_namespace} already contains an asset named '#{project_name}'"
         raise "Project name already taken"
       when /'asset.create' failed/
 
         # Ensure that the metadata validations are run
-        project.metadata_model.validate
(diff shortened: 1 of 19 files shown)


app/models/project_mediaflux.rb
```diff
@@ -33,16 +33,18 @@ def self.create!(project:, session_id:, xml_namespace: nil)
...
         raise "Project name already taken"
       when /'asset.create' failed/
 
         # Ensure that the metadata validations are run
-        project.metadata_model.validate
-        raise TigerData::MissingMetadata.missing_metadata(schema_version:"0.6", errors: project.metadata_model.errors)
+        if project.valid? &&  project.metadata_model.valid?
+          raise response_error[:message]  # something strange went wrong
+        else
+          raise TigerData::MissingMetadata.missing_metadata(schema_version:"0.6", errors: project.metadata_model.errors)
```
bess, line 46: Should this use `TigerdataSchema::SCHEMA_VERSION` instead of hard coding the version number?

spec/models/project_mediaflux_spec.rb
```diff
@@ -103,6 +101,20 @@
           expect(incomplete_project.metadata_model.errors.attribute_names[0].to_s).to eq "project_id"
         end
       end
+
+      it "should raise a error if any error occurs in mediaflux" do
```
bess, line 105: 😍

bess (approved):
This looks great! I made one suggestion but I wouldn't block merging over it.

app/models/project_mediaflux.rb
carolyncole: Likely.  That was not the code change I was working on, but I got tired of not seeing the real error from Mediaflux.  I'll update it.

bess (approved):

PR #672, opened 2024-05-01
Updating ProjectMediaflux to work with Rubocop

carolyncole:
I noticed the class was formatted realy weird.  I believe the @param and the param were confusing rubocop.  As soon as I changed `update_metadata(params: params)` (notice the second param) then rubocop reformatted the class

### DIFF
--- a/app/models/project_metadata.rb
+++ b/app/models/project_metadata.rb
@@ -18,7 +18,7 @@ class ProjectMetadata
   def initialize(project:, current_user: nil)
     @project = project
     @current_user = current_user
-    @params = { }
+    @params = {}
   end
 
   # Generates a Hash of updated Project metadata attributes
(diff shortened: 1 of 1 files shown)


app/models/project_metadata.rb
```diff
@@ -33,42 +33,37 @@ def activate_project(collection_id:)
...
+    # create two provenance events, one for approving the project and another for changing the status of the project
     project.provenance_events.create(event_type: ProvenanceEvent::APPROVAL_EVENT_TYPE, event_person: current_user.uid, event_details: "Approved by #{current_user.display_name_safe}")
     project.provenance_events.create(event_type: ProvenanceEvent::STATUS_UPDATE_EVENT_TYPE, event_person: current_user.uid, event_details: "The Status of this project has been set to approved")
   end
-  
-  def create( params:)
-    project.metadata = update_metadata(params:)
+
+  def create(params:)
+    project.metadata = update_metadata(params: params)
```
carolyncole, line 64: This is the change that started rubocop working again

hectorcorrea: 🎉

hectorcorrea (approved):

PR #681, opened 2024-05-06
Moving project parsing into the AssetMetadataRequest

carolyncole:


spec/models/project_metadata_spec.rb
```diff
@@ -229,5 +217,39 @@
         expect(activate_event.event_details).to eq "Activated by Tigerdata Staff"
       end
     end
+
+    # check the logic for not activating when the
```
bess, line 221: Looks like this comment wasn't finished?

PR #689, opened 2024-05-07
creating a background job to activate projects

### LINKED ISSUE #630: Background Job to check Approved Projects in Mediflux
Then we need to have a job that goes out to mediaflux and confirms the collection exists and the project id (DOI) matches.
The status in our system gets updated to Active

Completion criteria:

- [x] Happens in the background
- [x] Contacts mediaflux to query for the project id (DOI) based on the collection id
- [x] send and email to the system administrator who approved the project if the project id does not match
- [x] send a notification to Honey badger that indicates the miss match occured
- [x] Updates the system status to Active only if the project id matches.  The title no longer shows (pending) to the user

JaymeeH:
closes #630


bess (approved):
Very nice! I appreciate the use of comments. Makes it easier to follow the logic, and long-term makes the code easier to maintain. Thank you.

PR #703, opened 2024-05-09
Make the project_directory parent path the default path if none is set

### LINKED ISSUE #702: Approve Screen has a . for the project directory prefix
### Expected behavior
It should have the project_directory_parent_path

### Actual behavior
instead it has a `.`

carolyncole:
fixes #702


JaymeeH (approved):
Thanks! This fixed the parent path issue

Issue #684, opened 2024-05-07
When I create a project via ProjectMediaflux.create! Accumulators should be created


carolyncole:
### Acceptance criteria

- [ ] ProjectMediaflux.create! should create the Accumulators that [the TestProjectGenerator creates}(https://github.com/pulibrary/tiger-data-app/blob/main/app/services/test_project_generator.rb#L17-L20)
- [ ] The TestProjectGenerator should no longer call the accumulator creations separately

PR #709, opened 2024-05-10
Refactor ProjectMediaflux tests

bess:
By actually connecting to MediaFlux we catch some errors that we were otherwise missing. Also, it'll make it easier to extend this class without needing to re-create our mocked interactions every time.

This test suite tests for the same set of conditions, but without mocked mediaflux interactions, and with more meaningful checks in several cases.

Ref #684

### DIFF
--- a/spec/models/project_mediaflux_spec.rb
+++ b/spec/models/project_mediaflux_spec.rb
@@ -1,80 +1,92 @@
 # frozen_string_literal: true
 require "rails_helper"
 
-RSpec.describe ProjectMediaflux, type: :model, stub_mediaflux: true do
-  let(:namespace_request) { instance_double(Mediaflux::Http::NamespaceCreateRequest, resolve: true, "error?": false) }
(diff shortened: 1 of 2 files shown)


spec/models/project_mediaflux_spec.rb
```diff
@@ -1,80 +1,92 @@
...
-          expect(mediaflux_id).to be "123"
+        it "creates a project namespace" do
+          namespace_path = mediaflux_metadata[:namespace]
+          namespace_metadata = Mediaflux::Http::NamespaceDescribeRequest.new(
+            session_token: current_user.mediaflux_session, 
+            path: namespace_path
+            ).metadata
+          
+          expect(namespace_metadata[:description]).to match(/Namespace for/)
+        end
```
carolyncole, line 78: @bess since this is now talking to mediflux, could we move all these tests into 1 it block? Aka 3 expects in the test instead of 3 tests?

carolyncole (changes requested):
@bess just one small request to make the test suite a bit faster.

carolyncole (approved):
We decided to leave as is for clarity...

PR #710, opened 2024-05-10
Adding a view for devise so we can prompt the user to login correctly

carolyncole:
refs #677


leefaisonr (approved):
This looks good. Thank you!

PR #712, opened 2024-05-10
Raise an error when attempting to destroy an invalid namespace

### LINKED ISSUE #692: Namespace Destroy can silently fail
### Expected behavior
We should see errors that occur when the test suite can not clean up.  We should either raise an error in the [NamespaceDestroy](https://github.com/pulibrary/tiger-data-app/blob/main/app/models/mediaflux/http/namespace_destroy_request.rb#L22) or the code that calls the destroy if there is an issue

### Actual behavior
At the moment any error just gets ignored and test fail for no apparent reason

JaymeeH:
- raising an error when attempting to destroy a namespace that does not exist
- testing namespace destroy on an invalid namespace
- allowing connect_to_mediaflux support spec to continue despite encountering a namespace destroy error
closes #692

### DIFF
--- a/spec/models/project_mediaflux_spec.rb
+++ b/spec/models/project_mediaflux_spec.rb
@@ -59,14 +59,6 @@ RSpec.describe ProjectMediaflux, type: :model do
         end
         let(:project_parent) { Rails.configuration.mediaflux["api_root_collection"] }
 
-        before do
-          # Destroy the parent collection
-          Mediaflux::Http::NamespaceDestroyRequest.new(
-            session_token: current_user.mediaflux_session,
-            namespace: project_parent
-          ).destroy
-        end
-
         it "creates a project namespace" do
           namespace_path = mediaflux_metadata[:namespace]
           namespace_metadata = Mediaflux::Http::NamespaceDescribeRequest.new(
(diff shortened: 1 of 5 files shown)


spec/models/project_mediaflux_spec.rb
```diff
@@ -59,14 +59,6 @@
...
         let(:project_parent) { Rails.configuration.mediaflux["api_root_collection"] }
 
-        before do
-          # Destroy the parent collection
-          Mediaflux::Http::NamespaceDestroyRequest.new(
-            session_token: current_user.mediaflux_session,
-            namespace: project_parent
-          ).destroy
-        end
-
```
bess, line 69: Why did you remove this? This test section is "when the parent collection does not exist", so ensuring that the parent collection does not exist before the tests run is important.

bess (changes requested):
Overall this looks great, but I have a couple of questions.

spec/models/project_mediaflux_spec.rb
JaymeeH: I deleted it because I was under the impression that whenever the test runs, it automatically deletes everything in mediaflux by default since it has the `connect_to_mediaflux:` tag at the describe level so I assumed it was redundant and attempting to delete the namespace twice. Maybe i'm wrong about that?

spec/models/project_mediaflux_spec.rb
```diff
@@ -138,8 +130,7 @@
         expect {
           ProjectMediaflux.create!(project: incomplete_project, session_id: session_token)
         }.to raise_error do |error|
-          expect(error).to be_a(StandardError)
-          expect(error.message).to include("call to service 'asset.create' failed: XPath tigerdata:project/ProjectID is invalid: missing value")
+          expect([MediafluxDuplicateNamespaceError, StandardError, RuntimeError]).to include(error.class)
```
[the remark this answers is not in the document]
JaymeeH: Yes theoretically, but looking at this now — a duplicate namespace should never occur in this test. I can update it to be more clear.

bess (approved):

PR #713, opened 2024-05-10
Add accumulators when we create a mediaflux project

### LINKED ISSUE #684: When I create a project via ProjectMediaflux.create! Accumulators should be created

bess:
Fixes #684

### DIFF
--- a/app/assets/stylesheets/_settings.scss
+++ b/app/assets/stylesheets/_settings.scss
@@ -298,3 +298,8 @@
     min-height: 38px;
   }
 }
+
+.login button {
+  background-color: #ffc107;
+  border-color: #ffc107;
+}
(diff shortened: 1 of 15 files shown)


carolyncole (approved):
@bess I might suggest removing the save! and the assigning of the mediaflux id from here https://github.com/pulibrary/tiger-data-app/blob/main/app/models/project.rb#L101-L102 Is that part of another PR?

PR #715, opened 2024-05-13
Initializing a quota for a project when it is created in mediaflux

### LINKED ISSUE #685: Project Quotas should get generated when ProjectMediaflux.create! gets called
### Acceptance criteria

- [x] [CreateCollectionQuotaRequest](https://github.com/pulibrary/tiger-data-app/blob/main/app/models/mediaflux/http/create_collection_quota_request.rb) gets called with the correct values in ProjectMediaflux.create!
- [x] Please make this its own method, like [this one](https://github.com/pulibrary/tiger-data-app/blob/main/app/models/project_mediaflux.rb#L59-L78). The create! method is far to big and unwieldy.

JaymeeH:
closes #685

### DIFF
--- a/app/models/project_mediaflux.rb
+++ b/app/models/project_mediaflux.rb
@@ -53,6 +53,7 @@ class ProjectMediaflux
     project.mediaflux_id = id
     project.save!
     self.create_accumulators(mediaflux_project_id: id, session_id: session_id)
+    self.create_quota(project: project, mediaflux_project_id: id, session_id: session_id)
     id
   end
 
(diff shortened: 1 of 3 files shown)


spec/models/project_mediaflux_spec.rb
```diff
@@ -50,6 +49,18 @@
...
 
+      describe "quota", connect_to_mediaflux: true do
+        it "adds a quota when it creates a project in mediaflux" do
+          project = FactoryBot.create(:project_with_doi)
+          described_class.create!(project: project, session_id: current_user.mediaflux_session)
+          metadata = Mediaflux::Http::AssetMetadataRequest.new(
+            session_token: current_user.mediaflux_session, 
+            id: project.mediaflux_id
+            ).metadata
+          expect(metadata[:quota_allocation]).not_to be_empty
```
carolyncole, line 60: @JaymeeH could you check for `500 GB` here? ( I think that is the default allocation)

app/models/project_mediaflux.rb
```diff
@@ -77,6 +78,17 @@ def self.create_accumulators(mediaflux_project_id:, session_id:)
     accum_size.resolve
   end
 
+  def self.create_quota(project:, mediaflux_project_id:, session_id:)
+    #TODO: SHOULD WE CREATE A PROJECT USING REQUESTED VALUES OR APPROVED VALUES?
```
carolyncole, line 82: Good question.  Should we bring this up to the team?  No approval step happens, so I would assume the requested values like you did here.

carolyncole (approved):
This can be merged as is, but I would love if the test check for an actual value.

app/models/project_mediaflux.rb
JaymeeH: Yea I think it's a good idea to bring this up during the next stand-up

Issue #704, opened 2024-05-09
Move status to top of project approval page and add flash highlighting


kelynch:
Move the "status" value out of the provenance section of the page and put it near the top with [Bootstrap flash highlighting](https://getbootstrap.com/docs/4.0/components/alerts/) or something similar.  See current state and requested changes mockup below: 

Current example:
![Screenshot 2024-05-08 at 10 10 21 AM](https://github.com/pulibrary/tiger-data-app/assets/7795797/829eb02b-9e8a-41ed-bd6d-ba52e39313fa)

Request mockup: 
![Screenshot 2024-05-09 at 11 49 25 AM](https://github.com/pulibrary/tiger-data-app/assets/7795797/768fcb5a-d034-428b-8bc8-5ad6e31c0680)

## Acceptance criteria
- [ ] The "status" field is at the top of the page, underneath the project name and "new project in the queue" text blurb.
- [ ] The "status" field is highlighted with "alert" when pending and "success" when approved.

PR #717, opened 2024-05-14
Display status banner at top of page

leefaisonr:
Ref #704

### DIFF
--- a/app/views/projects/show.html.erb
+++ b/app/views/projects/show.html.erb
@@ -10,6 +10,21 @@ Project Details:
   </div>
 <% end %>
 
+<% if !@project.in_mediaflux? %>
+  <div class="alert alert-warning" role="alert">
+    <dl>
+    <dt>Status</dt> <dd><%=@project_status%></dd>
+    </dl>
+  </div>
+<% end %>
+<% if @project.in_mediaflux? %>
+  <div class="alert alert-success" role="alert">
+    <dl>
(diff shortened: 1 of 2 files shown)


app/views/projects/show.html.erb
```diff
@@ -10,6 +10,21 @@ Project Details:
...
 <% end %>
 
+<% if !@project.in_mediaflux? %>
+  <div class="alert alert-warning" role="alert">
+    <dl>
+    <dt>Status</dt> <dd><%=@project_status%></dd>
+    </dl>
+  </div>
+<% end %>
+<% if @project.in_mediaflux? %>
```
hectorcorrea, line 20: This could be an `<% else %>` too, right?

app/views/projects/show.html.erb
```diff
@@ -10,6 +10,20 @@ Project Details:
   </div>
 <% end %>
 
+<% if !@project.in_mediaflux? %>
+  <div class="alert alert-warning" role="alert">
+    <dl>
+    <dt>Status</dt> <dd><%=@project_status%></dd>
+    </dl>
+  </div>
+<% else @project.in_mediaflux? %>
```
hectorcorrea, line 19: It can be just `<% else %>` - you don't need the call to `@project.in_mediaflux?` here.

leefaisonr: Ahh, correcting now - thanks!

PR #719, opened 2024-05-15
Status highlight displays based on project status

JaymeeH:
Refactoring the status highlight to display based on the projects current status, instead of whether or not it is in mediaflux.


carolyncole (commented):
@JaymeeH the change looks ok, but what is this changing?  Can you write a test to show why you are changing this?  Aka what is it fixing?

carolyncole (approved):
Thanks @JaymeeH

Issue #618, opened 2024-03-28
Establish the Tester-Trainer role (as defined by the EPIC #554)


carolyncole:
Our User needs to have the concept of the trainer role.
https://github.com/pulibrary/tiger-data-app/issues/554

Completion Criteria:
- [x] Update the User Model to include the ability to know is a user is `trainer?`
- [x] Update the rake task to load this role when loading in users

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/618?planning-poker) with Zenhub @bess @JaymeeH @jrgriffiniii @leefaisonr @kelynch

PR #728, opened 2024-05-21
Add tester trainer role

leefaisonr:
Ref #618


app/models/user.rb
```diff
@@ -105,6 +105,8 @@ def eligible_sysadmin?
     return true if superuser || sysadmin
   end
 
+  def tester_trainer; end
```
carolyncole, line 108: Was this an intended method?

carolyncole (commented):
Looks good!  Only one question (see comments)

app/models/user.rb
leefaisonr: Thanks @carolyncole for the suggested change

carolyncole (approved):
Thanks!

PR #730, opened 2024-05-21
Creating a MediafluxTime class to handle timezone inconsistencies

### LINKED ISSUE #568: Re-think time zone approach
When we wrote our timezone ADR we thought that mediaflux ALWAYS reported time in UTC, and did not include timezone data on the dates it sends us. This is not the case. Mediaflux uses UTC for the internal system time but will report a timestamp based on the user's session. For example, here is Robert Knight querying aterm and getting a `US/Pacific` timezone:

Screenshot 2024-03-13 at 2 42 25 PM

I believe we should be checking the timezone when we get dates from mediaflux, and adjusting the timezone and formatting in a consistent way across the application. Maybe a `MediafluxTime` class that takes a whole snip from mediaflux and gives back an America/New_York and iso time format? 

Screenshot 2024-03-13 at 2 32 30 PM

JaymeeH:
Taking into consideration the the fact that an asset may have a timezone attached to it before converting it to local Princeton time.

- Converts time to `UTC` using the `gmt-offset` attached to the field
- Converts time to `America/New_York` before formatting to `iso8601` 

closes #568

### DIFF
--- a/app/models/mediaflux_time.rb
+++ b/app/models/mediaflux_time.rb
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 class MediafluxTime
+  # Converts the givin time snippet into local princeton time while accounting for potential time zone of the asset.
+  #
+  # @param xml_snip [:nokigiri], response_xml from a mediaflux request
+  # @return [String] returns a iso8601 Princeton time value.
   def convert(xml_snip:)
-    # takes an xml snip of a nokigiri based time object from a mediaflux response and returns a Princeton time value
-    # xml_snip: (required) The xml path to a time object from mediaflux
-
     xml = xml_snip
     time = xml.text.to_time
     gmt = xml.xpath("./@gmt-offset").text.to_f
--- a/app/models/mediaflux_time.rb
+++ b/app/models/mediaflux_time.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 class MediafluxTime
   def convert(xml_snip:)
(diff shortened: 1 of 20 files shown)


app/models/mediaflux_time.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+class MediafluxTime
+  def convert(xml_snip:)
+    # takes an xml snip of a nokigiri based time object from a mediaflux response and returns a Princeton time value
```
bess, line 4: I *love* these code comments. How would you feel about putting them in YARD format so they'll show up nicely in our auto-built documentation? I think if we can get in the habit of doing that it'll make this project more sustainable long-term. 
Here is an example of a YARD method header:
```
# Converts the object into textual markup given a specific format.
#
# @param format [Symbol] the format type, `:text` or `:html`
# @return [String] the object converted into the expected format.
def to_format(format = :html)
  # format the object
end
```

bess (approved):
@JaymeeH This looks great and is a much better approach than what we were doing previously. Thank you! I left one comment, about re-writing your code comments to match what YARD expects (so they'll show up nicely formatted [here](https://pulibrary.github.io/tiger-data-app/)) but I wouldn't block a merge for that.

app/models/mediaflux_time.rb
JaymeeH: Will do! ☝🏿

Issue #714, opened 2024-05-10
Mediaflux projects are being created at the root instead of in /td-test-001


bess:
### Expected behavior
I expect that when I run the automated tests, it will create mediaflux objects in /td-test-001

### Actual behavior
Mediaflux objects are being created at the root
<img width="346" alt="Screenshot 2024-05-10 at 4 37 53 PM" src="https://github.com/pulibrary/tiger-data-app/assets/65608/a907fda7-8c83-4c00-8ce2-1ca9aec342be">

### Steps to replicate
Run the test suite locally and look at mediaflux in the browser

### Impact of this bug
Tests fail when we get namespace collisions, because our test cleanup isn't working, because it is expecting everything to be in /td-test-001

bess:
These are from a test that ensures that an error is raised when a project is submitted with missing required metadata. I switched to testing it against mediaflux instead of against mocked interactions and discovered that although the error is raised, mediaflux still makes an asset, in an unexpected place.

carolyncole:
@bess the name space is made in a step before the collection is created.  Unsure why the namespace is not getting created under tigerdata.  That said I know there is no code to clean that up.  So mediaflux is not creating the collection even when fields are missing.

bess:
@carolyncole Oh, that's helpful thank you!

PR #732, opened 2024-05-23
Bugfix: Ensure test directories are valid paths

bess:
Ref #714


JaymeeH (approved):
Looks good, Thank you!

PR #738, opened 2024-05-24
Adding the ability to call to mediaflux to push an update to the submission event

### LINKED ISSUE #729: Write Provenance to Mediaflux
### User story

We should be able to send an API call to write a provenance event to mediaflux.  We will utilize the asset.update to write an event to mediaflux.

This can then be used when we are allowed to write provenance events to mediaflux directly.  It can also be utilized with xtoshell to write the TCL script for writing a provenance event that can be called for a safe write of provenance to mediaflux.

### Acceptance criteria

- [ ] A class exists that can update an asset with a provenance event
- [ ] The class creates an xtoshell_xml payload that can be utilized in script.execute

carolyncole:
Unfortunately we have to replace the entire document we can not just replace single elements. 

fixes #729

### DIFF
--- a/app/models/mediaflux/http/asset_approve_request.rb
+++ b/app/models/mediaflux/http/asset_approve_request.rb
@@ -1,0 +1,170 @@
+# frozen_string_literal: true
+module Mediaflux
+  module Http
+    # Constructs a request to mediaflux to approve a project
+    #
+    # @example
+    #  project = Project.first
+    #  project.save_in_mediaflux(session_id: User.first.mediaflux_session)
+    #  approve_req = Mediaflux::Http::AssetApproveRequest.new(session_token: User.first.mediaflux_session, project:)
+    #  approve_req.resolve
+    #
+    class AssetApproveRequest < Request
+      attr_reader :project_metadata, :project
+      # Constructor
+      # @param session_token [String] the API token for the authenticated session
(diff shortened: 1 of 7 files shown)


app/models/mediaflux/http/asset_approve_request.rb
```diff
@@ -0,0 +1,170 @@
+# frozen_string_literal: true
+module Mediaflux
+  module Http
+    # Constructs a request to mediaflux to approve a project
+    #
+    # @example
```
bess, line 6: 😍 Loving the YARD docs, thank you.

app/models/mediaflux/http/request.rb
```diff
@@ -142,6 +142,13 @@ def xtoshell_xml( name: self.class.service)
         xml.strip.gsub("\"","'").gsub("<args>","").gsub("</args>","")
       end
 
+      # This method is used for transforming iso8601 dates to dates that MediaFlux likes
```
bess, line 145: Don't we have a method that does this already? https://github.com/pulibrary/tiger-data-app/blob/main/app/models/mediaflux_time.rb#L24-L30

bess (approved):
LGTM, except that I think we already have a method for formatting mediaflux dates?

PR #746, opened 2024-05-29
Upgrading yarn for deployment to servers

carolyncole:
Also adding .yarn directory and file

Ref #776


carolyncole:
I believe this is no longer needed based on all the recent update

Issue #648, opened 2024-04-18
Refactor MediaFlux HTTP Requests to name them consistently 


leefaisonr:
Some of our MediaFlux requests libraries are named `noun_verb_request.rb`, like `namespace_create_request.rb`. Others are named `verb_noun_request.rb` like `create_asset_request.rb`. We should pick one pattern and stick to it. The `noun_verb` pattern better matches what MediaFlux does. 

Acceptance Criteria: 

- [x] In `app/models/mediaflux/http` all requests libraries are consistent with naming
- [x] In `spec/models/mediaflux/http` all requests libraries are consistent with naming

PR #749, opened 2024-05-30
update create http request names

Seanwarren77:
ref #648


carolyncole (approved):
Thank you @Seanwarren77  and @leefaisonr

PR #750, opened 2024-05-30
542 cleaning exported files

### LINKED ISSUE #542: Cleaning exported files
With PR https://github.com/pulibrary/tiger-data-app/pull/530 we are going to start saving files generated by the user to a shared location, which allows users to download them regardless of what server produced the file (see https://github.com/pulibrary/tiger-data-app/issues/523)

We should come up with a process to decide when to delete those files:
* Should we delete them as soon as the user downloads the file?
* Should we delete them after X number of days (regardless of whether the user downloads them or not)?
* Other ideas

PR https://github.com/pulibrary/tiger-data-app/pull/530 added a rake task to delete these files, but we need to run the rake task manually. 

Completion Criteria
- [x] Keep it for one week at the longest
- [x] Keep it until I have done two more further file inventories.  Aka they requested the complete list two more times.
- [x] Remove the download link from recent activity that has been deleted

JaymeeH:
Adding whenever gem to tigerdata
Automated cron job deletes exported files that are more than one week old.

closes #542

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -1,0 +1,24 @@
+# Use this file to easily define all of your cron jobs.
+#
+# It's helpful, but not entirely necessary to understand cron before proceeding.
+# http://en.wikipedia.org/wiki/Cron
+
+# Example:
+#
+# set :output, "/path/to/my/cron_log.log"
+#
+# every 2.hours do
+#   command "/usr/bin/some_great_command"
+#   runner "MyModel.some_method"
+#   rake "some:great:rake:task"
+# end
+#
+# every 4.days do
+#   runner "AnotherModel.prune_old_records"
+# end
+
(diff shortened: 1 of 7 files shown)


config/schedule.rb
```diff
@@ -0,0 +1,25 @@
...
+#   rake "some:great:rake:task"
+# end
+#
+# every 4.days do
+#   runner "AnotherModel.prune_old_records"
+# end
+
+# Learn more: http://github.com/javan/whenever
+
+every 1.week do
```
carolyncole, line 23: I might make this run daily, so that there are not some files that stay around for 13 days.
```suggestion
every 1.day do
```

carolyncole (commented):
@JaymeeH I do not think this closes the ticket.  I do not see any code that  `Removes the download link from recent activity that has been deleted`

Issue #753, opened 2024-05-31
Remove the concept of tiger_data_values from ProjectMediaflux and Mediaflux::Http


carolyncole:
### What maintenance needs to be done?

It is confusing to have a hash being passed around that is not really tested.  We have a bunch of places where the Schema version was not updated and it would simplify ProjectMediaflux to pass a project object to MediaFlux::Http::AssetCreateRequest and MediaFlux::Http::AssetUpdateRequest.  They could share code with MediaFlux::Http::AssetApproveRequest, which puts data from a project into the xml format that mediaflux needs.

The concept of changing the data for mediaflux into a hash and then changing it to XML adds a layer of complexity that is not needed and is hard to debug.

### Level of urgency
- [x] High
- [ ] Moderate
- [ ] Low

### Acceptance criteria
 - [x] Convert the test for MediaFlux::Http::AssetUpdateRequest to contact Mediaflux directly (connect_to_mediaflux: true)
 - [ ] Convert the test for MediaFlux::Http::AssetCreateRequest to contact Mediaflux directly (connect_to_mediaflux: true)
 - [x] MediaFlux::Http::AssetCreateRequest accepts a project instead of a hash
 - [ ] MediaFlux::Http::AssetUpdateRequest accepts a project instead of a hash
 - [ ] ProjectMediaflux no longer creates project_values, which is currently passed to the above classes as tigerdata_value

PR #757, opened 2024-06-05
Updating AssetCreateRequest to only create collections

carolyncole:
We really are not creating anything but collection assets and likely will not be in the future. No need to keep the additional code options.

Also changed one of the tests to contact mediaflux, which required updates to AssetMetadataRequest to validate all the input.

refs #753


leefaisonr (approved):
This looks great, thanks @carolyncole !

PR #760, opened 2024-06-07
Splitting project creation into a specialized ProjectCreate class

carolyncole:
Remove project metadata from asset create
Also utilize the project instead of a hash when initializing the project creation class refs #753


JaymeeH (approved):
Thank you, looks good

Issue #773, opened 2024-06-13
Rename UserJob to UserRequest 


leefaisonr:
Right now we have a model called UserJob. This is confusing, because its named as "job" which implies that it is a background job. It isn't, it's a record in the database of requests made by a user. Because this is a rename of a model, create migrations to ensure that we can deploy this to a server without deleting the database. 

## Implementation notes 

https://marouenbousnina.com/tutorials/2018-06-11-rename-activerecord-model/

bess:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/773?planning-poker) with Zenhub @carolyncole @hectorcorrea @JaymeeH @jrgriffiniii @leefaisonr

PR #780, opened 2024-06-17
Tigerdata Refactor: User Requests

JaymeeH:
Refactoring tigerdata User Jobs to be User Requests. 

refs #773


bess (approved):
This refactor was done as an ensemble.

PR #796, opened 2024-06-26
Display Mediaflux version on the footer 

### LINKED ISSUE #790: Add the mediaflux version to the footer
### User story
As a developer, I want to know exactly what version of mediaflux the system is running against. 

Call https://github.com/pulibrary/tiger-data-app/blob/main/app/models/mediaflux/http/version_request.rb

### Acceptance criteria

- [ ] The version of mediaflux displays in the footer, near the version of tiger-data-app

hectorcorrea:
Display Mediaflux version on the footer. I am only displaying it on the homepage since getting the Mediaflux version requires an extra call to Mediaflux and we don't want to do that for every single page.

Screenshot 2024-06-26 at 3 18 02 PM

Closes #790

### DIFF
--- a/app/views/welcome/index.html.erb
+++ b/app/views/welcome/index.html.erb
@@ -97,3 +97,10 @@
     </div>
     <% end %>
   </div>
+
+<script type="module">
+  // Make the AJAX call for version information explicitly on this page
+  // (and for now only on this page) so that we don't make extra calls
+  // to Mediaflux for every page loaded.
+  displayMediafluxVersion();
+</script>
--- a/config/mediaflux.yml
+++ b/config/mediaflux.yml
@@ -53,7 +53,7 @@ test:
   api_root_collection_namespace: '/td-test-001'
   api_root_collection: 'path=/td-test-001/tigerdata'
   api_transport: 'http'
-  api_host: '0.0.0.0'
+  api_host: 'mediaflux.example.com'
   api_port: '8888'
   api_domain: 'system'
   api_user: 'manager'
--- a/config/mediaflux.yml
+++ b/config/mediaflux.yml
@@ -53,7 +53,7 @@ test:
   api_root_collection_namespace: '/td-test-001'
   api_root_collection: 'path=/td-test-001/tigerdata'
   api_transport: 'http'
-  api_host: 'mediaflux.example.com'
+  api_host: '0.0.0.0'
   api_port: '8888'
   api_domain: 'system'
   api_user: 'manager'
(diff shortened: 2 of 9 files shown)


config/mediaflux.yml
```diff
@@ -53,7 +53,7 @@ test:
   api_root_collection_namespace: '/td-test-001'
   api_root_collection: 'path=/td-test-001/tigerdata'
   api_transport: 'http'
-  api_host: 'mediaflux.example.com'
+  api_host: '0.0.0.0'
```
hectorcorrea, line 56: Darn...this breaks a ton of tests. Need to figure out how to un-stub only certain calls to Mediaflux.

bess:
@hectorcorrea What if we made an initializer that grabs the mediaflux version once at system start and it sets an environment variable? Then we we wouldn't have to use javascript for this.

config/mediaflux.yml
carolyncole: @hectorcorrea 'connect_to_mediaflux: true' on any tests un stubs just that one

spec/system/welcome_spec.rb
```diff
@@ -59,6 +59,14 @@
         expect(page).not_to have_content "Please log in"
         expect(page).to have_content "Log Out"
       end
+
+      it "shows the Mediflux version on the home page for a logged in user" do
```
carolyncole, line 63: This would change to:
```suggestion
      it "shows the Mediflux version on the home page for a logged in user", connect_to_mediaflux: true do
```

config/mediaflux.yml
carolyncole, line 56:
```suggestion
  api_host: 'mediaflux.example.com'
```

carolyncole (commented):
I think you will be in business with my two changes @hectorcorrea

hectorcorrea:
@carolyncole that indeed took care of the problem. I added the `stub_mediaflux: true` to a couple of other specs that load the home page since that where the version is displayed.

app/views/welcome/index.html.erb
```diff
@@ -96,4 +96,11 @@
...
-  </div>
\ No newline at end of file
+  </div>
+
+<script type="module">
+  // Make the AJAX call for version information explicitly on this page
+  // (and for now only on this page) so that we don't make extra calls
+  // to Mediaflux for every page loaded.
+  displayMediafluxVersion();
+</script>
```
hectorcorrea, line 106: This is really neat...by using `type="module"` we can execute a function defined in `application.js`

(the function is in fact defined in `mediafluxVersion.js` but it is imported and made available via `application.js`)

carolyncole (approved):
I wonder if we should be caching this information vs calling on every page.  Certainly good enough for the moment.  Thanks @hectorcorrea!

bess (approved):

Issue #768, opened 2024-06-12
Create a class named ProjectAccumulators that knows to create three accumulators for a project


carolyncole:
### What maintenance needs to be done?
Create a class in either the models directory or the models/mediaflux directory based on the implementors feeling.  The code for this ticket already [exists here]( https://github.com/pulibrary/tiger-data-app/blob/main/app/models/project_mediaflux.rb#L63-L78)

```
my_project = Project.new; ProjectAccumulator.new(my_project).create!
ProjectAccumulator should be idempotent - we should be able to run this more than once and it should just do the right thing
ProjectAccumulator.new(my_project).validate => check that existing accumulators match our expectation
```

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Acceptance criteria
 - [x] Pull the code from ProjectMediaflux create_accumulators into a class
 - [x] the class will have two methods one for create!, which does nothing if nothing is needed and validate, which determines if the project has accumulators correctly applied.
 - [x] Write a test that connects to mediaflux and creates the accumulators on a project
 - [x] Write a test that connects to mediaflux and validates the accumulators on a project

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/768?planning-poker) with Zenhub @hectorcorrea @JaymeeH @jrgriffiniii @kelynch @leefaisonr

JaymeeH:
@carolyncole what does `which does nothing if nothing is needed ` mean in this context? we want ProjectAccumulators.create! to actually create all three necessary accumulators right?

carolyncole:
@JaymeeH the `create!` method (`one for create!, which does nothing if nothing is needed`)

JaymeeH:
Ok — I understand

JaymeeH:
@carolyncole right now, so far we have only ever created accumulators for `collection.asset.count` and `content.all.size`, do you know what the third accumulator is supposed to be used for?

carolyncole:
This has the information near the end on what accumulators Chuck creates https://princetonrc.slack.com/files/U05T4KHDH4P/F077BRM891P/project_creation_script_for_pena-ffcws.txt

PR #797, opened 2024-06-26
refactoring project accumulators

JaymeeH:
Creating a ProjectAccumulators class that creates and validates all of the necessary accumulators for a project

refs #768

### DIFF
--- a/app/models/project_accumulator.rb
+++ b/app/models/project_accumulator.rb
@@ -1,16 +1,16 @@
 # frozen_string_literal: true
 class ProjectAccumulator
-    def initialize(project:, session_id:)
-        @collection_id = project.mediaflux_id
-        @session_id = session_id
-    end
+  def initialize(project:, session_id:)
+    @collection_id = project.mediaflux_id
+    @session_id = session_id
+  end
 
-    # Create accumulators for all newly created mediaflux projects
+  # Create accumulators for all newly created mediaflux projects
   #
   # @param mediaflux_project_id [] the id of the project that needs accumulators
   # @param session_id [] the session id for the user who is currently authenticated to MediaFlux
-  def create!()
-    accum_names = validate()
+  def create!
+    accum_names = validate
     return true if accum_names == true
 
     # Create accumulators if they do not exist
@@ -25,10 +25,10 @@ class ProjectAccumulator
     end
   end
 
-  def validate()
+  def validate
     collection_metadata = Mediaflux::Http::AssetMetadataRequest.new(session_token: @session_id, id: @collection_id).metadata
     accum_names = collection_metadata[:accum_names].to_a.map(&:to_s)
-    (accum_names.size == 3) ? true : accum_names
+    accum_names.size == 3 ? true : accum_names
   end
 
   private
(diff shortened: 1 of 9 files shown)


bess (changes requested):
@JaymeeH The overall logic on this looks really good, but I'm confused about why ProjectAccumulator has a database table. It doesn't look like we're storing anything there, right? I think you have a few options for how to resolve that... I'd be curious about whether this still works if we just remove the database migration, since I can't see that it is being used anywhere. If I've missed something please let me know!

app/models/project_accumulator.rb
```diff
@@ -0,0 +1,65 @@
...
+  def initialize(project:, session_id:)
+    @collection_id = project.mediaflux_id
+    @session_id = session_id
+  end
+
+  # Create accumulators for all newly created mediaflux projects
+  #
+  def create!
+    accum_names = validate
+    return true if accum_names == true
```
carolyncole, line 12: What does the true signify here?  I might have expected true to signify a change and false to signify no change.  What does the method return of a change has occurred?

spec/models/project_accumulator_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      expect(metadata[:accum_names].map(&:to_s)).to include("accum-size")
+      expect(metadata[:accum_names].map(&:to_s)).to include("accum-store-size")
+    end
+
+    it "does not create accumulators if they already exist" do
+      ProjectMediaflux.create!(session_id: user.mediaflux_session, project: project_in_mediaflux)
+      project_accumulators = described_class.new(project: project_in_mediaflux, session_id: session_id)
+      project_accumulators.create!
+
+      expect(project_accumulators.create!).to eq(true)
```
carolyncole, line 26: I'm a bit confused what the true means in this context, see my comment above.

spec/models/project_accumulator_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+RSpec.describe ProjectAccumulator, connect_to_mediaflux: true, type: :model do
+  let(:user) { FactoryBot.create(:user) }
+  let(:session_id) { user.mediaflux_session }
+  let(:project_in_mediaflux) { FactoryBot.create(:project_with_doi, status: Project::APPROVED_STATUS) }
+
+  describe "#create!" do
+    it "creates accumulators for the project" do
+      project_id = ProjectMediaflux.create!(session_id: user.mediaflux_session, project: project_in_mediaflux)
+      project_accumulators = described_class.new(project: project_in_mediaflux, session_id: session_id)
+      project_accumulators.create!
```
carolyncole, line 13: What is the return value here.  Since the return value is significant below maybe it should be tested here

carolyncole (changes requested):
Hi @JaymeeH nice work!  Just one little source of confusion. (see coments)

app/models/project_accumulator.rb
JaymeeH: @carolyncole  Would it make more sense if `#create!` returned true on a change and false if no change occurs?

bess (approved):
🌟

carolyncole (approved):
Thanks @JaymeeH

PR #803, opened 2024-06-28
Adds class to create the root collection asset

### LINKED ISSUE #791: Create a class for  MediafluxRootCollectionAsset
### User story
A `MediafluxRootCollectionAsset` is a top level asset that all collections get created into. It's not a `Project` collection, it is the `root` of the Mediaflux system. Each project would be a "branch" off of this. 

### Acceptance criteria

- [ ] Calling `MediafluxRootCollectionAsset.create!` should ensure the root/trunk asset exists, and create it if it does not
- [ ] `asset_create_request.rb` is the class that currently does this, but it doesn't have a `#create!` method
- [ ] The `#create!` method should be idempotent -- if it exists already, do not recreate it

### Implementation notes, if any

Screenshot 2024-06-26 at 11 37 14 AM

hectorcorrea:
Added two new `Mediaflux:http` classes: one to check if an asset exists and another to create the root collection asset. 
Added a `Mediaflux::RootCollectionAsset` class to wrap the logic to create the root collection asset (checks and all).

I have not updated the code to switch the places where we create the root collection since I wanted to keep this PR small. That will come as a separate PR.

Part of #791


app/models/mediaflux/root_collection_asset.rb
```diff
@@ -0,0 +1,24 @@
...
+      @error = nil
+    end
+
+    def create
+      check_root = Mediaflux::Http::AssetExistRequest.new(session_token: @session_token, path: @path)
+      return true if check_root.exist?
+
+      create_request = Mediaflux::Http::CollectionAssetCreateRootRequest.new(session_token: @session_token, namespace: @namespace, name: @name)
+      create_request.resolve
+      @error = create_request.response_error
```
carolyncole, line 20: Nice!

carolyncole (approved):
Thanks @hectorcorrea.  I assume there is a follow on PR to remove asset.create and call this class in the code?

hectorcorrea:
@carolyncole yes, I am planning on replacing the existing calls to create the root collection as a separate PR.

PR #808, opened 2024-07-02
Create root namespace as part of RootCollectionAsset

hectorcorrea:
Updated the `RootCollectionAsset` class to take care of the root namespace as well. Added a new class to check whether a namespace exists too.

(code to create the root has not yet been integrated into the existing code, but the functionality to create the root collection and its associated namespace is now complete and will be integrated in the next PR)

Part of #791

### DIFF
--- a/app/models/mediaflux/root_collection_asset.rb
+++ b/app/models/mediaflux/root_collection_asset.rb
@@ -5,12 +5,14 @@ module Mediaflux
 
     def initialize(session_token:, namespace:, name:)
       @session_token = session_token
-      @namespace = namespace || Rails.configuration.mediaflux["api_root_collection_namespace"]   # /td-prod-001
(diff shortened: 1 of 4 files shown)


app/models/mediaflux/root_collection_asset.rb
```diff
@@ -5,20 +5,38 @@ class RootCollectionAsset
...
+    private
+
+      def create_root_namespace
+        check_root_namespace = Mediaflux::Http::NamespaceExistRequest.new(session_token: @session_token, namespace: @namespace)
+        return true if check_root_namespace.exist?
+
+        create_namespace_request = Mediaflux::Http::NamespaceCreateRequest.new(session_token: @session_token, namespace: @namespace)
+        create_namespace_request.resolve
+        @error = create_namespace_request.response_error
+        @error.blank?
```
hectorcorrea, line 39: This is the new code to make sure the root namespace exists before we create the root collection.

jrgriffiniii (approved):

Issue #769, opened 2024-06-12
Convert ProjectMetadata class to have ruby attributes instead of being based on a hash


carolyncole:
### What maintenance needs to be done?

Utilizing a hash is hard to test and unreliable.  We would like the class to explicitly know about all attributes needed and be able to validate those attributes.

We also want the Metadata class to be able to be used to generate the json for the project.to_json and the jsonb for the database save.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Acceptance criteria
 - [x] Add a [save_to_mediaflux method](https://github.com/pulibrary/tiger-data-app/blob/main/app/models/project_mediaflux.rb#L92-L93) `ProjectMetadata.new(my_project).save_to_mediaflux`
 - [x] Add ruby class properties for the metadata json values of the project
 - [x] Add a test that contacts mediaflux to replace the metadata for a project that has a mediaflux id by calling `save_to_mediaflux`
 - [x] All other methods will not be touched
 - [x] Clean up the code to call save_to_mdiaflux instead of ProjectMediaflux#update and remove the method in ProjectMediaflux

carolyncole:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/769?planning-poker) with Zenhub @hectorcorrea @JaymeeH @jrgriffiniii @kelynch @leefaisonr

PR #811, opened 2024-07-02
Replacing hash values in ProjectCreateRequest 

hectorcorrea:
Started prototyping how to replace the hash values in ProjectCreateRequest with explicit values in ProjectMetadata

Moves #769 forward


JaymeeH (approved):
👍🏿 Part 1 of this refactor

PR #816, opened 2024-07-03
ProjectMetadata Refactor continued

### LINKED ISSUE #769: Convert ProjectMetadata class to have ruby attributes instead of being based on a hash

JaymeeH:
Refactoring project#metadata to be a Metadata class instead of a hash
Co-authored-by: Hector Correa <hectorcorrea@users.noreply.github.com>

closes #769 

The main changes in this PR are:
* `ProjectMetadata` class is now a Plain Old Ruby Object (PORO) with no dependencies to the `Project` model or `Mediaflux`. 
* Use this new `ProjectMetadata` PORO instead of a hash in the `Project` model and everywhere where we need the project metadata...which was a lot of places.
* Move some of the functionality from `ProjectMetadata` to `Project` (e.g. `approve` and `activate` now live on the `Project` class rather than the `ProjectMetadata` class)


app/controllers/projects_controller.rb
```diff
@@ -67,18 +69,15 @@ def create
   def show
     project
     @departments = project.departments.join(", ")
-    @project_metadata = project.metadata
+    @project_metadata = project.metadata_model
 
-    sponsor_uid = @project_metadata[:data_sponsor]
-    @data_sponsor = User.find_by(uid: sponsor_uid)
+    @data_sponsor = User.find_by(uid: @project_metadata.data_sponsor)
```
carolyncole, line 74: I do not think this line is needed twice
```suggestion
    @data_sponsor = User.find_by(uid: @project_metadata.data_sponsor)
```

app/controllers/projects_controller.rb
```diff
@@ -104,48 +103,27 @@ def show
 
   def edit
     project
-    if project.metadata[:status] != Project::APPROVED_STATUS
+    if project.metadata_model.status != Project::APPROVED_STATUS
       flash[:notice] = "Pending projects can not be edited."
       redirect_to project
-    elsif project.metadata[:status] == Project::APPROVED_STATUS && !eligible_editor? #check if the current user is a sponsor of manager
+    elsif project.metadata_model.status == Project::APPROVED_STATUS && !eligible_editor? #check if the current user is a sponsor of manager
```
carolyncole, line 109: This is not really part of your PR, but since you are touching this line might as well fix the typo
```suggestion
    elsif project.metadata_model.status == Project::APPROVED_STATUS && !eligible_editor? #check if the current user is a sponsor or a manager
```

app/models/project.rb
```diff
@@ -2,55 +2,118 @@
...
-    @metadata_model ||= ProjectMetadata.new(project: self)
+    @metadata_model ||= ProjectMetadata.new_from_hash(self.metadata)
+  end
+
+  def metadata_model=(new_metadata_model)
+    @metadata_model = new_metadata_model
   end
 
-  def metadata=(metadata)
-    metadata[:schema_version] ||= TigerdataSchema::SCHEMA_VERSION
```
carolyncole, line 23: Where is this PR does the current Schema version get defaulted?

app/models/project.rb
```diff
@@ -95,29 +158,17 @@ def self.approved_projects
   def self.data_user_projects(user)
     # See https://scalegrid.io/blog/using-jsonb-in-postgresql-how-to-effectively-store-index-json-data-in-postgresql/
     # for information on the @> operator
-    query_ro = '{"data_user_read_only":["' + user + '"]}'
-    query_rw = '{"data_user_read_write":["' + user + '"]}'
+    query_ro = '{"ro_users":["' + user + '"]}'
```
carolyncole, line 161: This is changing what is being store in the database.  This means you need a migration for all existing projects, or we need to clear all projects.  What made you decide to store the information in the database differently?

app/models/project.rb
```diff
@@ -242,18 +293,12 @@ def files_from_iterator(iterator_resp)
       lines
     end
 
-    def metadata_with_defaults
-      data = (metadata_json&.dup || { }).with_indifferent_access
-      Rails.configuration.project_defaults.each { |key, value| data[key] ||= value }
```
carolyncole, line 247: Where do the defaults get included with the metadata in this PR? `Rails.configuration.project_defaults` pulls in the config file and inserts the default values

app/models/project_mediaflux.rb
```diff
@@ -6,6 +6,22 @@ class MediafluxDuplicateNamespaceError < StandardError
 
 # Take an instance of Project and adds it to MediaFlux
 class ProjectMediaflux
+  # If the project hasn't yet been created in mediaflux, create it.
```
carolyncole, line 9: Nice!

app/models/project_metadata.rb
```diff
@@ -1,125 +1,120 @@
...
+    @project_purpose = metadata_hash[:project_purpose]
+    @project_directory = metadata_hash[:project_directory]
+
+    @storage_capacity = metadata_hash[:storage_capacity]
+    @storage_performance_expectations = metadata_hash[:storage_performance_expectations]
+    @created_on = metadata_hash[:created_on]
+    @created_by = metadata_hash[:created_by]
+    @updated_by = metadata_hash[:updated_by]
+    @updated_on = metadata_hash[:updated_on]
+    set_defaults
```
carolyncole, line 46: Using this version of the defaults will cause a regression.  We need to be utilizing the `Rails.configuration.project_defaults` here

app/models/project_metadata.rb
```diff
@@ -1,125 +1,120 @@
...
+    @project_directory = params[:project_directory]
+
+    @storage_capacity = params[:storage_capacity]
+    @storage_performance_expectations = params[:storage_performance_expectations]
+
+    @created_by = params[:created_by] if params[:created_by]
+    @created_on = params[:created_on] if params[:created_on]
+    @updated_by = params[:updated_by] if params[:updated_by]
+    @updated_on = params[:updated_on] if params[:updated_on]
+    set_defaults
```
carolyncole, line 73: Using this version of the defaults will cause a regression. We need to be utilizing the Rails.configuration.project_defaults here

app/models/project_metadata.rb
```diff
@@ -130,39 +125,60 @@ def user_list_params(counter, key_prefix)
...
-          timestamps[:created_on] = Time.current.in_time_zone("America/New_York").iso8601
-
-        else
-          timestamps[:created_by] = project.metadata[:created_by]
-          timestamps[:created_on] = project.metadata[:created_on]
-          timestamps[:updated_by] = current_user.uid
-          timestamps[:updated_on] = Time.current.in_time_zone("America/New_York").iso8601
+      # Initializes values that we have defaults for.
+      def set_defaults
+        if @storage_capacity.nil?
```
carolyncole, line 130: Using this version of the defaults will cause a regression. We need to be utilizing the Rails.configuration.project_defaults here

app/models/project_metadata.rb
```diff
@@ -130,39 +125,60 @@ def user_list_params(counter, key_prefix)
...
       end
 
-      def form_metadata
-        ro_users = user_list_params(read_only_counter, "ro_user_")
-        rw_users = user_list_params(read_write_counter, "rw_user_")
-        data_users = {
-          data_user_read_only: ro_users,
-          data_user_read_write: rw_users
+      # Sets a value in the object if the value exists in the params
+      def set_value(params, key)
```
carolyncole, line 145: Is this going to stop people from blanking out values in the future?  Aka if the param value is empty I can not set it to blank.

app/models/project.rb
```diff
@@ -2,55 +2,118 @@
...
   APPROVED_STATUS = "approved"
   ACTIVE_STATUS = "active"
 
-  delegate :to_json, to: :metadata_json
+  delegate :to_json, to: :metadata_json # field in the database
+
+  def create!(initial_metadata:, user:)
+    self.metadata_model = initial_metadata
+    if initial_metadata.project_id.blank?
+        self.draft_doi(user: user)
```
carolyncole, line 21: We should check to see if the project is valid before the DOI is drafted.  You are setting the DOI (which calls out to datacite) and then not saving that doi if the item is not valid.

app/models/project.rb
```diff
@@ -2,55 +2,118 @@
...
+    self.metadata_model.project_directory = mediaflux_metadata[:project_directory]
+    self.save!
+
+    ProvenanceEvent.generate_active_events(project: self, user: current_user)
+  end
+
+  def draft_doi(user:)
+    puldatacite = PULDatacite.new
+    self.metadata_model.project_id = puldatacite.draft_doi
+    # self.save!
```
carolyncole, line 67: You are not saving the doi to the database if the project is not valid, which means you will be drafting many many DOIs that are not saved if the project does not validate.

app/models/project_validator.rb
```diff
@@ -1,12 +1,15 @@
...
         # Required fields, always validate
-        validate_role(project:, netid: project.metadata[:data_manager], role: "Data Manager")
-        validate_role(project:, netid: project.metadata[:data_sponsor], role: "Data Sponsor")
+        validate_role(project:, netid: project.metadata_model.data_manager, role: "Data Manager")
+        validate_role(project:, netid: project.metadata_model.data_sponsor, role: "Data Sponsor")
 
         # Validate if present
-        project.metadata[:data_user_read_only]&.each { |read_only| validate_role(project:, netid: read_only, role: "Data User Read Only")}
-        project.metadata[:data_user_read_write]&.each { |read_write| validate_role(project:, netid: read_write, role: "Data User Read Write")}
+        project.metadata_model.ro_users&.each { |read_only| validate_role(project:, netid: read_only, role: "Data User Read Only")}
```
carolyncole, line 11: You are changing the name of the database field here.  Was that intended.  If so you need a data migration

app/services/mediaflux_script_factory.rb
```diff
@@ -89,7 +81,6 @@ def build_create_script
             :StorageCapacity < :Size "#{project.metadata_json['storage_capacity']['size']['requested']}>" :Unit #{project.metadata_json['storage_capacity']['unit']['requested']}"
             :StoragePerformance "#{project.metadata_json['storage_performance_expectations']['requested']}"
             :ProjectPurpose "#{project.metadata_json['project_purpose']}"
-            :Submission < :RequestedBy "#{requested_by}" :RequestDateTime "#{requested_date}" >
```
carolyncole, line 92: Why remove the Submission from the mediaflux object?

carolyncole (changes requested):
@JaymeeH & @hectorcorrea.  This is a huge PR.  I reviewed the changes to the models, but not the specs.  The overall request for changes are:
1. Do not rename the data_user_read_only to ro_user (why: requires a database migration and is less easy to read)
1. Do not rename the data_user_read_write to rw_user (why: requires a database migration and is less easy to read)
1. Do not stop utilizing the  Rails.configuration.project_defaults (why: it is a feature that was requested and you are creating a regression)
1. a couple small typos

app/models/project_validator.rb
hectorcorrea: Good point. We kept getting confused with the names `data_user_read_only` vs `ro_users` (and the read-write versions) so we picked one and ran with it, but later on we noticed that the keys are saved to the database. I didn't think we needed to migrate any data at this point, but we might as well play it save. We'll update the code to use the original keys.

app/models/project.rb
hectorcorrea: Restored them.

app/models/project.rb
hectorcorrea: I've put back the code that reads the values from the config

app/models/project_metadata.rb
hectorcorrea: Good catch. Fixed.

app/models/project_metadata.rb
hectorcorrea: Good catch, we are not reading them from the config again

app/models/project_metadata.rb
hectorcorrea: Good catch, we are not reading them from the config again

app/services/mediaflux_script_factory.rb
hectorcorrea: Restored it

app/models/project_metadata.rb
hectorcorrea: Good catch. I've updated the code to read them from `Rails.configuration.project_defaults` again.

app/models/project.rb
hectorcorrea: Fixed. We restored the logic to only mint the DOI once. 

We ended up declaring a constant `ProjectMetadata::DOI_NOT_MINTED` rather than using an empty string for the check.

app/models/project.rb
```diff
@@ -2,55 +2,118 @@
...
   APPROVED_STATUS = "approved"
   ACTIVE_STATUS = "active"
 
-  delegate :to_json, to: :metadata_json
+  delegate :to_json, to: :metadata_json # field in the database
+
+  def create!(initial_metadata:, user:)
+    self.metadata_model = initial_metadata
+    if self.valid?
+      if initial_metadata.project_id == ProjectMetadata::DOI_NOT_MINTED
```
hectorcorrea, line 21: We are using a constant rather than blank but the logic is the same as before.

app/models/project.rb
hectorcorrea: Done.

app/models/project.rb
hectorcorrea: Done.

app/models/project_metadata.rb
```diff
@@ -1,125 +1,133 @@
...
-  attr_accessor :title, :description, :status, :data_sponsor, :data_manager, :departments, :ro_users, :rw_users, :created_on, :created_by, :project_id, :project_purpose, :storage_capacity,
-:storage_performance_expectations, :updated_by, :updated_on
-
-  def initialize(project:, current_user: nil)
-    @project = project
-    initialize_from_metadata
-    @current_user = current_user
-    @params = {}
-  end
+  DOI_NOT_MINTED = "DOI-NOT-MINTED"
```
carolyncole, line 3: Nice Change!

spec/factories/project.rb
```diff
@@ -14,23 +14,23 @@
...
-      storage_capacity { { size: { requested: 500 }, unit: { requested: "GB" } } }
-      storage_performance { { requested: "standard" } }
+      storage_capacity { { size: { requested: 500 }, unit: { requested: "GB" } }.with_indifferent_access }
+      storage_performance { { requested: "standard" }.with_indifferent_access }
       project_purpose { "research" }
       project_directory { "big-data" }
       schema_version { ::TigerdataSchema::SCHEMA_VERSION }
       approved_by { nil }
       approved_on { nil }
+      # TODO: remove submission from the project factory
```
carolyncole, line 24: @hectorcorrea Submission is part of the metadata I'm unsure why you would remove it https://docs.google.com/document/d/1RDCowxE2ulfkh0GrqTln8Tr9t5XcWOZ2U--zRpiwe48/edit#heading=h.jnv7e4pqqw3u

spec/mailers/tigerdata_spec.rb
```diff
@@ -47,9 +47,11 @@
     mail = ActionMailer::Base.deliveries.last
 
     project.reload
-    expect(project.metadata).to include("submission")
-    expect(project.metadata["submission"]).to include("requested_by")
-    expect(project.metadata["submission"]).to include("request_date_time")
+    # These fields are not currently, nor ever were a part of the project metadata
```
carolyncole, line 50: These really are part of the metadata standard see https://docs.google.com/document/d/1RDCowxE2ulfkh0GrqTln8Tr9t5XcWOZ2U--zRpiwe48/edit#heading=h.jnv7e4pqqw3u

carolyncole (changes requested):
@hectorcorrea & @JaymeeH take a look at the metadata schema https://docs.google.com/document/d/1RDCowxE2ulfkh0GrqTln8Tr9t5XcWOZ2U--zRpiwe48/edit#heading=h.jnv7e4pqqw3u

You are removing thins that are really part of the schema

spec/mailers/tigerdata_spec.rb
hectorcorrea: Good catch. I've restored them in the last commit https://github.com/pulibrary/tiger-data-app/pull/816/commits/44a62de63984ffb1eb2e46772427778c162879a2

carolyncole (approved):

Issue #835, opened 2024-07-15
Manual Mediaflux installation


bess:
Go through the installation instructions and manually build a mediaflux server

## Acceptance criteria
- [x] Get a copy of mediaflux (here: https://www.arcitecta.com/software/mf/4.16.047/mflux-dev_4.16.047_jvm_1.8.jar )
- [ ] Go through the installation guide and install it manually (licenses are attached to this ticket)
- [ ] Document the process in markdown to make it easier to automate

[Mediaflux Installation Guide 4.x.pdf](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK1RkQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--dd63031b360e9af2877621c0f925b33fba94a0aa/Mediaflux%20Installation%20Guide%204.x.pdf)

bess:
[mf_licenses.tar.gz](https://github.com/user-attachments/files/16237618/mf_licenses.tar.gz)

bess:
[Configuring authentication_AD binding for Mediaflux.docx](https://github.com/user-attachments/files/16254123/Configuring.authentication_AD.binding.for.Mediaflux.docx)
[Installation Steps for td-mf-db1.docx](https://github.com/user-attachments/files/16254124/Installation.Steps.for.td-mf-db1.docx)
[mediaflux.service.docx](https://github.com/user-attachments/files/16254125/mediaflux.service.docx)
[mediaflux.service.txt](https://github.com/user-attachments/files/16254126/mediaflux.service.txt)

PR #841, opened 2024-07-16
Manual mflux installation notes

bess:
Ref #835


leefaisonr (approved):
These look great - thanks!

Issue #828, opened 2024-07-09
rspec ./spec/models/mediaflux/http/asset_exist_request_spec.rb:15 is failing oftent in CI


carolyncole:
### What maintenance needs to be done?

We should take time to figure out why it is so flaky

It looks like there is more than one flaky test, we just noted that `collection_asset_root_request` also is failing intermittently. 

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
This is slowing down PRs

### Acceptance criteria
 - [ ] The test passes during CI constantly

hectorcorrea:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tiger-data-app/828?planning-poker) with Zenhub @carolyncole @JaymeeH @jrgriffiniii @kelynch @leefaisonr

carolyncole:
All the following went green after a rerun from failed:
rspec ./spec/models/mediaflux/project_update_request_spec.rb:57
rspec ./spec/models/mediaflux/http/asset_exist_request_spec.rb:15
rspec ./spec/models/mediaflux/http/collection_asset_create_root_request_spec.rb:9

bess:
Documenting more flakiness: 
![Screenshot 2024-07-18 at 1 15 09 PM](https://github.com/user-attachments/assets/90bcf3c7-6a47-42e9-8fac-ad9836176701)

carolyncole:
https://app.circleci.com/insights/github/pulibrary/tiger-data-app/workflows/build_accept_deploy/tests?branch=main contains all flaky test runs as defined by CI in the flaky section

hectorcorrea:
@carolyncole this is great information, I didn't know CircleCI kept track of flaky tests.

PR #843 (merged a couple of hours ago) should take care of `Mediaflux::Http::CollectionAssetCreateRootRequest` -- if that works we can probably try to apply a similar approach to the rest.

carolyncole:
@hectorcorrea  The root namespace is cleared out before each test.  You can see in a couple of the flaky tests that the namespace not existing is the reason for the failure.  I'm guessing that there is some sort of error during the deletion creation https://github.com/pulibrary/tiger-data-app/blob/main/spec/support/connect_to_mediaflux.rb#L16-L24

PR #843, opened 2024-07-18
Try to make the test less flaky

hectorcorrea:
The only way I can see this test failing *sometimes* is if `FFaker::InternetSE.company_name_single_word` is giving the same value for another test in the entire RSpec suite during the same run. 

In this PR I am forcing the value generated for this test to be unique. This should prevent the tests on `collection_asset_create_root_request_spec.rb` from being flaky on CI. Fingers crossed.

Moves #828 forward.


leefaisonr (approved):

carolyncole:
@hectorcorrea FYI from looking at the flaky tests that exists, a few are checking the response error and those two are indicating that the root namespace does not exist.

PR #847, opened 2024-07-18
HTTP custom headers for Mediaflux authentication

### LINKED ISSUE #834: Add an http header to the requests we send to mediaflux
Add an http header to the requests we send to mediaflux which indicates whether the user has authenticated via CAS.

## Acceptance criteria
- [ ] If a user has authenticated via CAS, there is a header in the http requests sent to mediaflux
  - [ ] the header includes the netid `TIGERDATA_NETID` - the user netid that is logged in
  - [ ] the header includes domain:princeton `TIGERDATA_DOMAIN` - Hardcoded to `princeton`
  - [ ] the header includes the CAS timeout data `TIGERDATA_TIMEOUT` - Follow the format from CAS
- [ ] No call sto mediaflux should be made if we are not logged in, but if this happens then
     - [ ] If a user has not authenticated via CAS that header does not exist or is empty
     - [ ] If a user's CAS session has expired, that header does not exist or is empty

See #836 for the Mediaflux component of this ticket.

hectorcorrea:
Updated the `Mediaflux::AssetExistRequest` to allow it to pass custom HTTP headers in the request to Mediaflux to test authentication. 

The main change is in the base `Mediaflux::Request` class but only the `Mediaflux::AssetExistRequest` has been updated to pass the custom HTTP parameters. 

Updated the Help page to use as playground to execute requests with the current logged in user to pass the custom HTTP parameters when the `http-headers` query string parameter is set to true: 

* http://localhost:3000/help?http-headers=true
* http://localhost:3000/help?http-headers=false (default)

Wireshark capture showing the HTTP headers:

Screenshot 2024-07-18 at 3 28 15 PM

Closes #834

### DIFF
--- a/app/controllers/welcome_controller.rb
+++ b/app/controllers/welcome_controller.rb
@@ -26,5 +26,21 @@ class WelcomeController < ApplicationController
     end
   end
 
-  def help; end
+  def help
+    # Piggybacking on this page to pass custom HTTP headers to Mediaflux
+    # in a very controlled scenario.
+    root_ns = Rails.configuration.mediaflux["api_root_collection_namespace"]
+    parent_collection = Rails.configuration.mediaflux["api_root_collection_name"]
+    @test_path = Pathname.new(root_ns).join(parent_collection)
+    @test_http_headers = false
+    if current_user != nil
+      @test_http_headers = params["http-headers"] == "true"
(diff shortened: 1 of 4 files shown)


bess (approved):
🎉

PR #855, opened 2024-07-22
Code clean up for the Root collection class 

hectorcorrea:
Made sure the code to create the root namespace and the required collection and namespaces is all funneled through the RootCollectionAsset class.

In a future PR I would like to rename the keys in our config file to match the keys in our code (root_ns, parent_collection, and parent_namespace) but we'll tackle that as a separate PR to keep things small.


spec/support/connect_to_mediaflux.rb
```diff
@@ -1,36 +1,32 @@
...
-      namespace: Rails.configuration.mediaflux[:api_root_to_clean]
-    )
-    create_request.resolve
-    if create_request.error?
-      puts "Error creating #{Rails.configuration.mediaflux[:api_root_to_clean]}: #{create_request.response_error}" # allow the message to show in CI output
-    end
-  end
+  # then create it and the project root collection and namespace so it exists for any tests
+  create_test_root_namespace(user)
+  ProjectMediaflux.create_root_tree(session_id: user.mediaflux_session)
```
hectorcorrea, line 17: Rubocop reformated this file so the changes look daunting, but the only change really is to call `ProjectMediaflux.create_root_tree` instead of `ProjectMediaflux.create_root_ns` and `ProjectMediaflux.prepare_parent_collection`

app/models/project_mediaflux.rb
```diff
@@ -88,30 +87,10 @@ def self.document(project:, xml_namespace: nil)
...
-                                                                        name: Mediaflux::Connection.root_collection_name)
-        raise "Can not create parent collection: #{create_parent_request.response_error}" if create_parent_request.error?
-      else
-        raise "Error finding parent collection (#{project_parent}) #{get_parent.response_error}"
-      end
-    end
+  def self.create_root_tree(session_id:)
+    root_ns = Rails.configuration.mediaflux["api_root_collection_namespace"]
+    parent_collection = Rails.configuration.mediaflux["api_root_collection_name"]
+    req = Mediaflux::RootCollectionAsset.new(session_token: session_id, root_ns:, parent_collection:)
```
hectorcorrea, line 93: This class has the logic to make sure the root and the required nodes below it exist (i.e. parent_collection and parent_namespace)

bess (approved):

bess:
Really nice addition of clarifying comments too, thank you!

PR #858, opened 2024-07-23
ProjectMediaflux#update now sets updated_on/updated_by

hectorcorrea:
Updated `ProjectMediaflux#update` to set the `updated_on` and `updated_by` with defaults when not provided in the metadata.

The change was small but I did change the method signature so I had to update a bunch of test to pass the `user` rather then the `session_id` and be able to calculate the defaults.

### DIFF
--- a/app/models/project_mediaflux.rb
+++ b/app/models/project_mediaflux.rb
@@ -9,16 +9,17 @@ class ProjectMediaflux
   # If the project hasn't yet been created in mediaflux, create it.
   # If it already exists, update it.
   # @return [String] the mediaflux id of the project
-  def self.save(project:, session_id:, xml_namespace: nil)
+  def self.save(project:, user:, xml_namespace: nil)
+    session_id = user.mediaflux_session
     if project.mediaflux_id.nil?
-      mediaflux_id = ProjectMediaflux.create!(project: project, session_id: session_id)
+      mediaflux_id = ProjectMediaflux.create!(project: project, user: user)
       ProjectAccumulator.new(project: project, session_id: session_id).create!()
-      project.reload
       Rails.logger.debug "Project #{project.id} has been created in MediaFlux (asset id #{mediaflux_id})"
     else
-      ProjectMediaflux.update(project: project, session_id: session_id)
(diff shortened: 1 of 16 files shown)


app/models/project_mediaflux.rb
```diff
@@ -63,14 +65,24 @@ def self.create!(project:, session_id:, xml_namespace: nil)
...
 
-  def self.update(project:, session_id:)
-    Mediaflux::ProjectUpdateRequest.new(session_token: session_id, project: project).resolve
+  def self.update(project:, user:)
+    session_id = user.mediaflux_session
+    project.metadata_model.updated_on ||= Time.current.in_time_zone("America/New_York").iso8601
+    project.metadata_model.updated_by ||= user.uid
+    update_request = Mediaflux::ProjectUpdateRequest.new(session_token: session_id, project: project)
+    update_request.resolve
+    raise update_request.response_error[:message] if update_request.error?
```
hectorcorrea, line 82: This is the main change in this PR (sets updated_on/by when needed)

spec/models/project_mediaflux_spec.rb
```diff
@@ -128,4 +121,26 @@
...
+    it "honors updated_on/by values when provided" do
+      updated_on = Time.current.in_time_zone("America/New_York").iso8601
+      project.metadata_model.updated_on = updated_on
+      project.metadata_model.updated_by = "user123"
+      described_class.update(project: project, user: current_user)
+      expect(project.metadata_model.updated_on).to eq updated_on
+      expect(project.metadata_model.updated_by).to eq "user123"
+    end
+  end
 end
```
hectorcorrea, line 146: Tests the new functionality

app/models/project_mediaflux.rb
```diff
@@ -63,14 +65,24 @@ def self.create!(project:, session_id:, xml_namespace: nil)
...
+
+    # Save the ActiveRecord to make sure the updated metadata is also saved in our PostgreSQL database
+    project.save!
```
hectorcorrea, line 85: This is the main change in this PR (sets updated_on/by when needed)

bess (approved):

PR #859, opened 2024-07-23
Show aterm create script on the Approval screen

hectorcorrea:
Refactored the `MediafluxScriptFactory` to use the new metadata and root values and then hook it to the Approval screen so that we can surface this information to people creating projects in Mediaflux (see screenshot below).

The script will probably need tweaking based on Chuck's feedback, but the general idea is there.

Screenshot 2024-07-24 at 1 24 20 PM

### DIFF
--- a/lib/tasks/projects.rake
+++ b/lib/tasks/projects.rake
@@ -73,15 +73,17 @@ namespace :projects do
     puts "Mediaflux asset #{asset_id} updated"
   end
 
-  desc "Outputs to a file an aterm script to create a project (namespace and collection) in Mediaflux"
+  desc "Outputs to a file an aterm script to create a project in Mediaflux"
   task :create_script, [:project_id] => [:environment] do |_, args|
     project_id = args[:project_id]
-    service = MediafluxScriptFactory.new(project_id: project_id)
-    script = service.build_create_script
-
+    project = Project.find(project_id)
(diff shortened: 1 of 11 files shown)


lib/tasks/projects.rake
```diff
@@ -73,15 +73,12 @@ namespace :projects do
...
     project_id = args[:project_id]
-    service = MediafluxScriptFactory.new(project_id: project_id)
-    script = service.build_create_script
-
-    file_name = "project_create_#{project.id}.txt"
-    puts "Saving script to #{file_name}"
-    File.write(file_name, script)
+    project = Project.find(project_id)
+    service = MediafluxScriptFactory.new(project: project)
+    puts service.aterm_script
```
carolyncole, line 81: I might rather this still create a file by default.  Is there a reason for this change?

carolyncole (commented):
Looks good to me, thanks @hectorcorrea.  One question, is @matthewjchandler happy with the button showing up for all users? Or did I miss a switch?

carolyncole (approved):
Still curious why the rake task no longer outputs a file, but if that is intended...

hectorcorrea:
My thought was that the rake task can be redirected it to a file `> myfile.txt` if I just output to the console (and for my purposes outputting to the console was convenient). I'll add the file option again in a separate PR.

PR #861, opened 2024-07-25
Mediaflux status long polling

hectorcorrea:
**Notice that this PR is a draft. If [James' work with WebSockets](https://github.com/pulibrary/tiger-data-app/pull/862) works in staging we won't need this PR, if it does not then this PR is a good backup plan.**

Uses JavaScript long polling to check the status of Mediaflux. We check the status every 2 seconds when a page is loaded, and the longer the page stays loaded we slow down the checks (every 10 seconds, and then every 30 seconds). We could tweak these settings as we wish.

Below is an example on how this looks in staging:

Screenshot 2024-07-26 at 11 09 23 AM


app/javascript/channels/index.js
```diff
@@ -1,4 +1,2 @@
 // Import all the channels to be used by Action Cable
 //
-
-import './mediaflux_channel';
```
hectorcorrea, line 4: This PR replaces the WebSockets/ActionCable implementation with a good old AJAX call with long polling.

app/views/shared/_header.html.erb
```diff
@@ -6,12 +6,12 @@
   </a>
       <div class="mediaflux">
         Mediaflux Status
-        <div class="mediaflux-status inactive"></div>
+        <div class="mediaflux-status unknown"></div>
```
hectorcorrea, line 9: Display a brown circle when we have not checked the status (e.g. for a not logged in user)

hectorcorrea:
Closing now that https://github.com/pulibrary/tiger-data-app/pull/862 is working in staging

PR #880, opened 2024-08-06
Ensuring that @project.metadata can be resolved within the view templates

### LINKED ISSUE #877: Handle cases where @project.metadata is nil in View templates
## Backtrace

    line 6 of [PROJECT_ROOT]/app/views/projects/_approved_details.html.erb: _app_views_projects__approved_details_html_erb__1108489002782064710_4927680
    line 83 of [PROJECT_ROOT]/app/views/projects/show.html.erb: _app_views_projects_show_html_erb___4146019773808734818_4927660
    line 21 of [PROJECT_ROOT]/app/controllers/application_controller.rb: mediaflux_session

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/113559/faults/110023590)

jrgriffiniii:
Resolves #877

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -91,7 +91,7 @@ class ProjectsController < ApplicationController
     @project_eligible_to_edit = true if project.status == Project::APPROVED_STATUS && eligible_editor?
 
     @project_metadata = @project.metadata
-    @project_id = @project_metadata[:project_id]
+    @project_id = @project_metadata[:project_id] || {}
     @storage_capacity = @project_metadata[:storage_capacity]
     @size = @storage_capacity[:size]
     @unit = @storage_capacity[:unit]


app/controllers/projects_controller.rb
```diff
@@ -90,6 +90,24 @@ def show
 
     @project_eligible_to_edit = true if project.status == Project::APPROVED_STATUS && eligible_editor?
 
+    @project_metadata = @project.metadata
+    @project_id = @project_metadata[:project_id]
```
carolyncole, line 94: Did you mean to default to an empty hash so the rest of the calls would not fail if the metadata was nil?
```suggestion
    @project_id = @project_metadata[:project_id] || { }
```

carolyncole (changes requested):
I wonder if this would be a good opportunity to utilize the ProjectMetadata object in the views.  https://github.com/pulibrary/tiger-data-app/blob/main/app/models/project_metadata.rb

Either way I think you atleast need the small change I suggested to avoid a nil pointer error

jrgriffiniii:
Thank you very much @carolyncole, this was the solution!

carolyncole (approved):

PR #888, opened 2024-08-19
Switching Mediaflux test instances depending on a environment variable

### LINKED ISSUE #887: Create a flag to switch between Mediaflux testing environments
Currently, both CircleCI and our local test suite runs against the CI-Ansible build for Mediaflux. We’re going to want to be able to test against our local Mediaflux. We should implement some environment variable that could flip the switch. The way we do RUN_IN_BROWSER=true , maybe we could also have a LOCAL_MEDIAFLUX=true

Acceptance Criteria:
- [ ] Add a flag that switches the `api_host` environment variable from `mflux-ci.lib.princeton.edu` to `0.0.0.0`
- [ ] Change the URI construction in `tiger-data-app/app/models/mediaflux/request.rb` to include `Connection.port` when the flag is set to local development

JaymeeH:
closes #887

### DIFF
--- a/app/models/mediaflux/request.rb
+++ b/app/models/mediaflux/request.rb
@@ -14,8 +14,13 @@ module Mediaflux
       end
 
       def self.uri
-        # This is the necessary URI for the Ansible build(s) of Mediaflux
-          URI("#{Connection.transport}://#{Connection.host}/#{request_path}")
+        # Setting the URI to the Ansible build or Docker build of mediaflux depending on the environment
+        if ENV["MFLUX_LOCAL"]
+          byebug
+          URI("#{Connection.transport}://#{Connection.host}:#{Connection.port}#{request_path}")
+        else
+          URI("#{Connection.transport}://#{Connection.host}#{request_path}")
+        end
       end
 
       # Constructs a new HTTP POST request for usage with the Mediaflux API
--- a/app/models/mediaflux/request.rb
+++ b/app/models/mediaflux/request.rb
@@ -15,11 +15,12 @@ module Mediaflux
 
       def self.uri
         # Setting the URI to the Ansible build or Docker build of mediaflux depending on the environment
-        if ENV["MFLUX_LOCAL"]
-          URI("#{Connection.transport}://#{Connection.host}:#{Connection.port}#{request_path}")
-        else
-          URI("#{Connection.transport}://#{Connection.host}#{request_path}")
-        end
+        # if ENV["MFLUX_LOCAL"]
+        #   URI("#{Connection.transport}://#{Connection.host}:#{Connection.port}#{request_path}")
+        # else
+        #   URI("#{Connection.transport}://#{Connection.host}#{request_path}")
+        # end
+        URI("#{Connection.transport}://#{Connection.host}:#{Connection.port}#{request_path}")
       end
 
       # Constructs a new HTTP POST request for usage with the Mediaflux API
--- a/app/models/mediaflux/request.rb
+++ b/app/models/mediaflux/request.rb
@@ -16,7 +16,6 @@ module Mediaflux
       def self.uri
         # Setting the URI to the Ansible build or Docker build of mediaflux depending on the environment
         if ENV["MFLUX_LOCAL"]
-          byebug
           URI("#{Connection.transport}://#{Connection.host}:#{Connection.port}#{request_path}")
         else
           URI("#{Connection.transport}://#{Connection.host}#{request_path}")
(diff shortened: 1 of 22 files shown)


JaymeeH:
When running the suite with `MFLUX_LOCAL=true`, There are 2 failing system specs in `tiger-data-app/spec/system/project_spec.rb` and 4 failing system specs in `tiger-data-app/spec/system/project_roles_spec.rb`. It is still trying to open a connection to the ansible build instead of the docker build of Mediaflux even though `connect_to_mediaflux` is changing the `api_host` to `0.0.0.0` at the start of the test. I presume that this bug is being caused by the mediaflux version request that is made when a user is re-directed to the homepage, but since these two system specs are not testing mediaflux functionality and the underlying version_request_spec passes in both environments — we can open this pr and address the bug in a separate ticket. Or if someone has a solution that stands out it can be merged in this pr

app/models/mediaflux/request.rb
```diff
@@ -14,8 +14,12 @@ def self.request_path
       end
 
       def self.uri
-        # This is the necessary URI for the Ansible build(s) of Mediaflux
-          URI("#{Connection.transport}://#{Connection.host}/#{request_path}")
+        # Setting the URI to the Ansible build or Docker build of mediaflux depending on the environment
+        if ENV["MFLUX_LOCAL"]
+          URI("#{Connection.transport}://#{Connection.host}:#{Connection.port}#{request_path}")
```
carolyncole, line 19: I wonder if the port could just be added.  It seems weird that we are not allowing mediaflux to be on a different port...

spec/models/mediaflux/store_list_request_spec.rb
```diff
@@ -2,10 +2,13 @@
 require "rails_helper"
 
 RSpec.describe Mediaflux::StoreListRequest, connect_to_mediaflux: true, type: :model do
-  let(:mediaflux_url) { "http://mflux-ci.lib.princeton.edu/__mflux_svc__" }
+  let(:mediaflux_url) { Mediaflux::Request.uri.to_s }
   let(:user) { FactoryBot.create(:user) }
   let(:approved_project) { FactoryBot.create(:approved_project) }
   let(:mediaflux_response) { "<?xml version=\"1.0\" encoding=\"UTF-8\" ?>\n<response><reply type=\"result\"><result></result></reply></response>" }
+  # Docker and Ansible responses are the same, but may be different in the future
+  let(:docker_response_store) { { id: "1", name: "db", tag: "", type: "database" } }
```
carolyncole, line 10: Is it me or do these lines match.  Why the switch here?

spec/models/mediaflux/version_request_spec.rb
```diff
@@ -5,13 +5,15 @@
   subject(:request) { described_class.new(session_token: session_token) }
   let(:session_token) { user.mediaflux_session }
   let(:user) { FactoryBot.create(:user) }
+  let(:docker_response) { "4.16.032" }
```
carolyncole, line 8: No change requested, just noting this is sad that we have two different versions here.

spec/system/welcome_spec.rb
```diff
@@ -23,6 +23,9 @@
     let(:other_user) { FactoryBot.create(:user, uid: "zz123") }
     let(:no_projects_user) { FactoryBot.create(:user, uid: "qw999") }
     let(:no_projects_sponsor) { FactoryBot.create(:project_sponsor, uid: "gg717") }
+    let(:docker_response) { "4.16.032" }
```
carolyncole, line 26:
```suggestion
    let(:docker_response) { "Mediaflux: 4.16.032" }
```

spec/system/welcome_spec.rb
```diff
@@ -23,6 +23,9 @@
...
+    let(:ansible_response) { "4.16.047" }
```
carolyncole, line 27:
```suggestion
    let(:ansible_response) { "Mediaflux: 4.16.047" }
```

carolyncole (commented):
@JaymeeH a couple questions and suggestions.  Nice work so far!

spec/models/mediaflux/store_list_request_spec.rb
JaymeeH: They are the same now, but I was leaving room for the potential that they may differ in the future. Maybe one build has something stored in the tag field? We could just go back to having one field for the store if that makes more sense

app/models/mediaflux/request.rb
JaymeeH: True — we can just change the port too when we want to run against docker. I can push an update shortly

carolyncole (approved):
Thanks @JaymeeH

PR #892, opened 2024-08-23
Move the switching to local mediaflux into the configuration file

### LINKED ISSUE #891: Calling CI server even if MFLUX_LOCAL is true
I wanted to be certain I was not going to impact other when I ran destructive commands, so I ran a little test to make sure MFLUX_LOCAL=true was capturing all calls to Mediaflux.
Turns out there are some Ajax calls for the version that happen after the test concludes that were still calling out to CI:
Screenshot 2024-08-23 at 8 44 59 AM

carolyncole:
fixes #891

Make this small change in main to see your system connecting to the server even when you have requested local
Screenshot 2024-08-23 at 8 48 15 AM

### DIFF
--- a/config/mediaflux.yml
+++ b/config/mediaflux.yml
@@ -95,8 +95,8 @@ test:
   api_root_collection_namespace: '/td-test-001/test'
   api_root_collection: 'path=/td-test-001/test/tigerdata'
   api_transport: 'http'
-  api_host: 'mflux-ci.lib.princeton.edu'
-  api_port: <%= ENV["TEST_MEDIAFLUX_PORT"] || '80' %>
+  api_host: <%= ENV["MFLUX_LOCAL"] ? "0.0.0.0" : 'mflux-ci.lib.princeton.edu' %>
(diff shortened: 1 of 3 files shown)


config/mediaflux.yml
```diff
@@ -95,8 +95,8 @@ test:
   api_root_collection_namespace: '/td-test-001/test'
   api_root_collection: 'path=/td-test-001/test/tigerdata'
   api_transport: 'http'
-  api_host: 'mflux-ci.lib.princeton.edu'
-  api_port: <%= ENV["TEST_MEDIAFLUX_PORT"] || '80' %>
+  api_host: <%= ENV["MFLUX_LOCAL"] ? "0.0.0.0" : 'mflux-ci.lib.princeton.edu' %>
+  api_port: <%= ENV["MFLUX_LOCAL"] ? "8888" : '80' %>
```
JaymeeH, line 99: This is much cleaner 
Thanks @carolyncole

JaymeeH (approved):

Issue #896, opened 2024-08-27
TigerData schema (XSD)


hectorcorrea:
Review the XSD created by Matt and Hale for the TigerData schema and how can we produce an XML document with the data that we have that follows it:
https://drive.google.com/file/d/1qrdPXwAt57uqPmHxh0Y86zac4RvtoJ2e/view?usp=drive_link

See also an example metadata record that conforms to the schema:
https://drive.google.com/file/d/1WCu9P1EWfFC16mjM73AI0gEKRCvbgANF/view?usp=drive_link

### Completion Criteria
- [x] Read through the XSD document to check for clarity of intent
- [x] Read through the example XML to further understand intentions for project metadata
- [x] Note any concerns with the above as comments on this ticket with an "at" to Matt
- [x] Spend approximately 1 hour discussing the technical aspects of the XSD with Matt and Halle
- [ ] If any action items for RDSS come out of the discussion, work with Matt to create new tickets

PR #899, opened 2024-08-29
First pass at validating an XML document against the TigerData XSD

hectorcorrea:
Adds a rake task showing how to validate the sample XML document provided by Matt in ticket #896 against the XSD provided on that same issue.

The rake task and the sample XML and XSD files are mostly for demonstration purposes and very likely this code (if useful) will be integrated in to the `app` folder, but for now this is mostly for demo and documentation purposes.

Run with: 
`bundle exec rake xml_schema:validate_example`

In addition included a `td_xml_schema_101.md` file with some general guidance on how to read the TigerData XML Schema.

### DIFF
--- a/lib/assets/xml.xsd
+++ b/lib/assets/xml.xsd
@@ -1,0 +1,287 @@
+<?xml version='1.0'?>
+<?xml-stylesheet href="../2008/09/xsd.xsl" type="text/xsl"?>
+<xs:schema targetNamespace="http://www.w3.org/XML/1998/namespace" 
+  xmlns:xs="http://www.w3.org/2001/XMLSchema" 
+  xmlns   ="http://www.w3.org/1999/xhtml"
+  xml:lang="en">
+
+ <xs:annotation>
+  <xs:documentation>
+   <div>
+    <h1>About the XML namespace</h1>
+
+    <div class="bodytext">
+     <p>
+      This schema document describes the XML namespace, in a form
+      suitable for import by other schema documents.
+     </p>
+     <p>
+      See <a href="http://www.w3.org/XML/1998/namespace.html">
+      http://www.w3.org/XML/1998/namespace.html</a> and
+      <a href="http://www.w3.org/TR/REC-xml">
+      http://www.w3.org/TR/REC-xml</a> for information 
+      about this namespace.
+     </p>
+     <p>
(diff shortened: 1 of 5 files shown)


lib/tasks/xml_schema.rake
```diff
@@ -0,0 +1,39 @@
...
+
+  # Replace the reference to `xml.xsd` from the xsd with a reference to our
+  # local copy of the file, otherwise the validation tends to fail because
+  # Nokogiri cannot load the original `xml.xsd` file from the W3C
+  # source. ¯\_(ツ)_/¯
+  # For more information see: https://stackoverflow.com/a/18527198/446681
+  def use_local_xml_xsd(xsd)
+    local_xml_xsd = Pathname.new("./lib/assets/xml.xsd").realpath.to_s
+    xsd.sub!("https://www.w3.org/2001/xml.xsd", "file://#{local_xml_xsd}")
+  end
```
hectorcorrea, line 37: `use_local_xml_xsd(xsd)` is a horrible hack but that's the only way I could get Nokogiri to validate the XML against the XSD successfully.

lib/assets/xml.xsd
carolyncole, line 1: @hectorcorrea what is this file?  Is it needed?

carolyncole (commented):
This is great work @hectorcorrea!  Just one question.

lib/assets/xml.xsd
```diff
@@ -0,0 +1,287 @@
+<?xml version='1.0'?>
```
carolyncole, line 1:
```suggestion
<!-- Allows us to utilize Nokogiri for the validation.  For some reason utilizing the copy at  https://www.w3.org/2001/xml.xsd causes issues.  See xml_schema rake task for more information -->
<?xml version='1.0'?>
```

carolyncole (changes requested):
Ok, so maybe we can add a comment,  just to link the two files up...

carolyncole (approved):
Thanks @hectorcorrea!  My partially caffeinated brain is playing the part of us a year from now...  When we think what is this file and delete it...

hectorcorrea:
@carolyncole agree, this is something we'll forget as soon as *squirrel* :)

Issue #921, opened 2024-09-18
[Tiger Data/staging] NoMethodError: undefined method `[]' for nil


bess:
This bug prevents project pages from rendering, and is blocking today's Tiger Data release.

## Backtrace

    line 96 of [PROJECT_ROOT]/app/controllers/projects_controller.rb: show
    line 21 of [PROJECT_ROOT]/app/controllers/application_controller.rb: mediaflux_session

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/113559/faults/110791705)

kelynch:
Closing in favor of https://github.com/pulibrary/tigerdata-app/issues/924 per Feature Refinement discussion this morning.

PR #922, opened 2024-09-18
Default to 0 if storage capacity is undefined

bess:
Ref #921

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -93,8 +93,8 @@ class ProjectsController < ApplicationController
     @project_metadata = @project.metadata
     @project_id = @project_metadata[:project_id] || {}
     @storage_capacity = @project_metadata[:storage_capacity]
-    @size = @storage_capacity[:size]
-    @unit = @storage_capacity[:unit]
+    @size = @storage_capacity[:size] || 0 # TODO: What should happen if @storage_capacity is nil? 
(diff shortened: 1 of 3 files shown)


app/controllers/projects_controller.rb
```diff
@@ -93,8 +93,8 @@ def show
     @project_metadata = @project.metadata
     @project_id = @project_metadata[:project_id] || {}
     @storage_capacity = @project_metadata[:storage_capacity]
-    @size = @storage_capacity[:size]
-    @unit = @storage_capacity[:unit]
+    @size = @storage_capacity[:size] || 0 # TODO: What should happen if @storage_capacity is nil?
```
carolyncole: @bess I doubt this will solve the error you are seeing.  The error indicates that the @storage_capacity is nil, not the values in the hash

app/controllers/projects_controller.rb
```diff
@@ -93,8 +93,8 @@ def show
     @project_metadata = @project.metadata
     @project_id = @project_metadata[:project_id] || {}
     @storage_capacity = @project_metadata[:storage_capacity]
```
carolyncole, line 95:
```suggestion
    @storage_capacity = @project_metadata[:storage_capacity] || { size: 0, unit: "" } # TODO: Should be the default
```

carolyncole (changes requested):
Not sure this will fix anything

PR #923, opened 2024-09-18
Set NodeJS release to 20.15.1

jrgriffiniii:
Deprecates https://github.com/pulibrary/tigerdata-app/pull/906, as https://pulmirror.princeton.edu/mirror/node/dist/latest-v20.x/ does not support any releases in the 20.17 series. Resolves https://github.com/pulibrary/tigerdata-app/issues/810

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ commands:
 
 jobs:
 
-  rspec:
+  test:
     working_directory: ~/tiger_data
     machine:
       image: ubuntu-2404:2024.05.1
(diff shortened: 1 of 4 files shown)


bess (changes requested):
@jrgriffiniii I know that usually we like a multi-stage CI build, but tiger data is using a machine build, not a docker build, and the multi-stage CI takes a *really long time*. In this case, the multi-stage build does not save us time, instead it COSTS us significant time. Please remove those changes?

bess (approved):

PR #933, opened 2024-09-20
Execute a Mediaflux TCL script from Ruby

hectorcorrea:
First pass at installing a TCL script in Mediaflux that then we can execute from within our Ruby application. 

At this point this is mostly a proof of concept to see if executing scripts stored in the server is better approach to submitting individual Mediaflux commands. 

This example is hard-coded to a `fileList.tcl` script that must be installed ahead of time (the script is included in the PR). 

To install the script run (you only need to run this once):

```
bundle exec rake projects:script_upload[hc8719]
```

Once installed the script can be run as follows as many times are you need and for any path that you want to:

```
bundle exec rake projects:script_file_list[hc8719,/td-demo-001/tigerdata/jul24-412]
```

The current script returns a list of files for the given path, but the important part is that it does it via the `fileList.tcl` script installed in Mediaflux rather than by submitting individual `asset.query` commands. In other words, the logic lives in the server rather than the client.

Part of https://github.com/pulibrary/tigerdata-app/issues/905


lib/assets/fileList.tcl
```diff
@@ -0,0 +1,52 @@
+# DESCRIPTION:
```
hectorcorrea, line 1: This is the TCL script, most if it is comments since this is new territory for me, but the instructions also show how to run the script outside of the Rails application which is really handy for testing and tweaking it.

lib/tasks/projects.rake
```diff
@@ -106,6 +106,50 @@ namespace :projects do
     puts "File list downloaded for project #{project_id}"
   end
 
+  desc "Uploads the fileList TCL script to Mediaflux and makes it executable"
```
hectorcorrea, line 109: Upload the script to Mediaflux (so that it can be run from the Rails app)

lib/tasks/projects.rake
```diff
@@ -106,6 +106,50 @@ namespace :projects do
...
+    if upload_request.error?
+      puts "Error creating script: #{upload_request.response_error[:message]}"
+    else
+      puts "Created asset: #{upload_request.result} at #{full_path}"
+      make_executable_request = Mediaflux::ScriptMakeExecutableRequest.new(session_token: user.mediaflux_session, path: full_path)
+      make_executable_request.resolve
+    end
+  end
+
+  desc "Runs the installed Mediaflux script to get the list of files under a given path"
```
hectorcorrea, line 132: Calls the script installed in Mediaflux.

carolyncole (approved):
Since this is a spike I'm ok with there not being tests...

PR #952, opened 2024-09-25
Only show the user the project directory levels that are visible to them

### LINKED ISSUE #941: Default namespace should show the user /tigerdata in production
the code should be able to know that the path is really princeton/tigerdata though since that is the real location, but tigerdata is what the user knows about

carolyncole:
Update the project presenter to show all the data vs allowing the show views to dig into the metadata and the metadata_model directly

fixes #941


app/models/mediaflux/connection.rb
```diff
@@ -62,5 +62,13 @@ def self.root_namespace
         Rails.configuration.mediaflux["api_root_ns"]
       end
     end
+
+    def self.hidden_root
```
jrgriffiniii, line 66: 👍

jrgriffiniii (approved):

PR #955, opened 2024-09-25
Making the edit page rout back to the project upon cancellation

### LINKED ISSUE #946: When you cancel from the Edit page, you should go to the project show
### User story

As a user I don't want to be bumped to the dashboard  when I cancel from the edit screen

### Acceptance criteria

- [ ] When a user hit cancel from the edit screen of a project they should go to the show page of the project

JaymeeH:
closes #946

### DIFF
--- a/app/assets/stylesheets/_settings.scss
+++ b/app/assets/stylesheets/_settings.scss
@@ -117,6 +117,17 @@
     margin-right: 23em;
     margin-left: 23.5em;
   }
+  .approved-details{
+    dt {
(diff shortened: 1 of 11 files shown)


carolyncole (commented):
It would be nice if there was a [test here](https://github.com/pulibrary/tigerdata-app/blob/main/spec/system/project_spec.rb#L135)

carolyncole (approved):
Thanks @JaymeeH

Issue #953, opened 2024-09-25
Don't display Mediaflux Id on Project Edit screen


hectorcorrea:
Issue #944 removed Mediaflux Id from the Show page but we should also remove it from the Edit screen:

![Screenshot 2024-09-25 at 10 12 59 AM](https://github.com/user-attachments/assets/30338390-bf31-4c75-8ae3-fd3a37af5fde)

PR #959, opened 2024-09-27
make mflux id only visible to superuser on edit pg

leefaisonr:
ref #953

### DIFF
--- a/app/views/projects/_edit_form.html.erb
+++ b/app/views/projects/_edit_form.html.erb
@@ -77,7 +77,7 @@
       </select>
     </div>
 
-    <% if @project.in_mediaflux? %>
+    <% if @project.in_mediaflux? && (current_user.superuser? || current_user.eligible_sysadmin?) %>
       <!--
         Make the field readonly so the user cannot change it, but leave it as an HTML INPUT so that it is
         send back to the controller (we don't want to lose this value)


app/views/projects/_edit_form.html.erb
```diff
@@ -77,7 +77,7 @@
       </select>
     </div>
 
-    <% if @project.in_mediaflux? %>
+    <% if @project.in_mediaflux? && (current_user.superuser? || current_user.eligible_sysadmin?) %>
```
carolyncole, line 80: @leefaisonr I think this may have unintended consequences of allow folks to edit the directory after the project has been saved to medaiflux. 
```suggestion
    <% if @project.in_mediaflux? %>
```

app/views/projects/_edit_form.html.erb
carolyncole, line 1: I would make the change to line 86 instead 
```
<% if (current_user.superuser? || current_user.eligible_sysadmin?) %>
   (MediaFlux id: <%= @project.mediaflux_id %>)<br/>
<% end %>

carolyncole (changes requested):
I think there will be side effects to this change.  I have suggested another...

app/views/projects/_edit_form.html.erb
leefaisonr: Thanks @carolyncole! I have made these requested changes.

carolyncole (approved):
Thanks @leefaisonr

PR #962, opened 2024-09-30
Document user and permissions when executing scripts

### LINKED ISSUE #905: Spike: Can we have a script in Mediaflux that allows our read only user to view a list as seen by an administrator
See if we can trigger a script to query the files in a collection and have it run as a higher permission

hectorcorrea:
Document how user's permissions work when executing scripts saved in Mediaflux

Closes #905


carolyncole (approved):
Thanks @hectorcorrea

PR #1024, opened 2024-11-05
Custom Error Handling

### LINKED ISSUE #1017: Adjust error handling behavior for accessibility
Current behavior:
Screenshot 2024-10-30 at 11 24 09 AM

Desired behavior:
Screenshot 2024-10-30 at 11 25 18 AM

[Full example screenshot here](https://private-user-images.githubusercontent.com/99462799/373792588-dce0d0b4-6abb-4eb1-b5f6-22709fe2e871.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzAzMDIwMzksIm5iZiI6MTczMDMwMTczOSwicGF0aCI6Ii85OTQ2Mjc5OS8zNzM3OTI1ODgtZGNlMGQwYjQtNmFiYi00ZWIxLWI1ZjYtMjI3MDlmZTJlODcxLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDEwMzAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQxMDMwVDE1MjIxOVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTljODJiYWNiNzJmNjAzYWQzZjBkY2YwOWZjZDM3NjUxNmIwZjZmYmU1ZjBlMTY0MmU2NGExZTk3YWQyNTMwOWUmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.aAJ6ijfF1G1fi8I3WLyM2XQDGkUMC6kdTNKgxnI0KaE)

## Acceptance criteria

- [x] The red border is added for error state
- [ ] The icon and text messages are implemented for error state
- [x] Ensure that the error is tagged with [aria-polite](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) and tested in a screenreader

JaymeeH:
closes #1017

### DIFF
--- a/app/views/projects/_edit_form.html.erb
+++ b/app/views/projects/_edit_form.html.erb
@@ -1,6 +1,6 @@
 <%= render 'form_errors' %>
 <%= render 'data_list' %>
-<div>
+<div id="edit">
    <h2>Project Roles</h2>
 
    <div class="row ">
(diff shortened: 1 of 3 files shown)


app/views/projects/_edit_form.html.erb
```diff
@@ -25,7 +33,15 @@
       </div>
       <div class="col-md-9">
          <% if !@project.persisted? || @project.metadata_model.data_sponsor == current_user.uid %>
-         <input type="text" list="manager-users" id="data_manager" aria-label="data manager" name="data_manager" required value="<%= @project.metadata_model.data_manager %>" /></input>
+         <input type="text" list="manager-users" id="data_manager" aria-label="data manager" name="data_manager" required oninvalid="this.setCustomValidity('')"
+               oninput="this.setCustomValidity('')" value="<%= @project.metadata_model.data_manager %>" /></input>
+         <div class="custom_error" id="manager_error" aria-live="polite">
+            <svg xmlns="http://www.w3.org/2000/svg" width="25" height="24" viewBox="0 0 25 24" fill="none">
```
carolyncole, line 39: I wonder if there is a way to extract this svg so it does not need to be in each location.  Assuming it is the same thing.   That said, this should not block the PR.

spec/models/mediaflux/time_spec.rb
```diff
@@ -23,7 +23,7 @@
         initial_tz = xml_snip.xpath("./@tz").text
 
         final_tz = instance.convert(xml_snip:)
-        expect(final_tz).to include("-04:00").or include("-05:00") #America/New_York changes based on daylights savings time
+        ["-04:00", "-05:00"].any? { |tz| final_tz.include?(tz) } #America/New_York changes based on daylights savings time
```
carolyncole, line 26: Does this really fail the test if the timezone is not one of the expected values?  I would assume it just returns false, which would not fail the test

spec/system/project_roles_spec.rb
```diff
@@ -29,10 +29,8 @@
     fill_in "data_manager", with: "xxx"
     page.find("body").click
     expect(page.find("button[value=Submit]")).to be_disabled
-    expect(page.find("#data_manager").native.attribute("validationMessage")).to eq "Please select a valid value."
```
carolyncole, line 32: How is the validation now being tested?

carolyncole (changes requested):
It looks like the changes to the automated tests may make things less tested.  Also if you can extract the SVG so it is not copied everywhere, that would be great!

spec/models/mediaflux/time_spec.rb
JaymeeH: Right, I can add an expectation check for the match to be true

spec/system/project_roles_spec.rb
JaymeeH: Through the button being disabled, but it would be ideal to test the css of the custom validation message

PR #1035, opened 2024-11-07
Started switching user data entry to use autocomplete

### LINKED ISSUE #980: Single Select User
**Single Select Design Update:**

The single select dropdown contains 5 stages -> Closed, Expanded, Filled, Result Not Found, and Error - Not Filled.

Closed: When the page loads, the form field appears in this state
Expanded: When the user clicks on this box, the drawdown list expands. The dropdown list contains a text-based search filter that helps users search for NetID from the list.
Filled: Once the user selects the desired NetID the dropdown closes and the selected value appears on the label.

image

Result Not Found: When the NetID is not present in the dropdown list, the dropdown indicates that no results were found.
Error - Not Filled: When a user attempts to submit the request on the New Project Request page without filling in a required field, the system will display an error message next to the missed field.

image

image

Figma Link:
https://www.figma.com/design/rgzEYTpnGl6l68moRSVwYg/Phase-1-UI-Details?node-id=190-1628&m=dev

_Originally posted by @gnemade51 in https://github.com/pulibrary/aux/issues/3#issuecomment-2394486778_

## Acceptance criteria

- [ ] Autocomplete functionality has been implemented for the data manager and data sponsor fields

hectorcorrea:
Updates the user entry elements on the form to use an autocomplete textbox and mimics the validation requested on #980.

This animated GIF shows both the autocomplete and the validation in action:
user-autocomplete

Closes #980


app/models/user.rb
```diff
@@ -19,27 +19,22 @@ def self.from_cas(access_token)
     user
   end
 
-  def self.all_users
-    User.all.map(&:uid)
-  end
-
+  # Users that can be project sponsors
   def self.sponsor_users
```
hectorcorrea, line 23: ` self.sponsor_users` and ` self.manager_users` below now returns a list of User objects (rather than just their `uid`s) so that we can display both the user name and their netid

spec/rails_helper.rb
```diff
@@ -88,3 +88,11 @@
     end
   end
 end
+
```
hectorcorrea, line 91: This helper removes the need of calling `page.find("body").click` as a separate step when validating textboxes since it forces the textbox to lose focus (via tabbing out)

app/views/projects/_edit_form.html.erb
```diff
@@ -95,120 +94,217 @@
...
-               <option><%= affiliation[:code] %></option>
-               <% end %>
-                <% end %>
-            </select>
-  </div>
-</div>
-</div>
+</div> <!-- id=edit -->
+
+<script>
```
hectorcorrea, line 191: This is the biggest change in this PR: the JavaScript logic to get the autocomplete and the validation to work together.

JaymeeH (approved):
Looks great!

public/jquery.autocomplete.js
```diff
@@ -0,0 +1,1044 @@
...
+})(
+    /**
+     * @param {jQuery} $
+     */
+    function ($) {
+    'use strict';
+
+    var utils = {
+            escapeRegExChars: function (value) {
+                return value.replace(/[|\\{}()[\]^$+*?.]/g, '\\$&');
```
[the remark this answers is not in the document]
hectorcorrea: This actually came from the plug-in so I didn't have to write it, thank goodness for that!

PR #1048, opened 2024-11-13
Add Capistrano task to remove application from the load balancer.

### LINKED ISSUE #1041: Add Capistrano task to remove application from the load balancer
## Acceptance Criteria

- [ ] Running `cap --hosts=<hosthere> staging application:remove_from_nginx` will prevent a node from serving traffic on the load balancer
- [ ] Running `cap --hosts=<hosthere> staging application:serve_from_nginx` will allow a node to serve traffic

## Technical Implementation

Add a health monitor that looks at a file on disk and reports a failure if it exists, so that nginx will fail its health check and pull it out.

Example PR: https://github.com/pulibrary/pdc_describe/pull/1985/files

We're really hoping `https://github.com/lbeder/health-monitor-rails/pull/127` will get merged.

## First Step

Update the health-monitor-rails gem so it has the new check.

tpendragon:
Closes #1041


config/initializers/health_monitor.rb
```diff
@@ -3,15 +3,20 @@
...
     # Mediaflux check
     config.add_custom_provider(MediafluxStatus)
 
     # Make this health check available at /health
     config.path = :health
 
     config.error_callback = proc do |e|
-      Rails.logger.error "Health check failed with: #{e.message}"
-      Honeybadger.notify(e)
+      unless e.is_a?(HealthMonitor::Providers::FileAbsenceException)
```
maxkadel, line 16: Same as Geaccirc - It occurred to me after I did this on another app that we probably do still want it to log to the Rails logger, in case we're like "why the heck is this not on the load balancer?"

maxkadel (approved):
Thanks @tpendragon!

Issue #1030, opened 2024-11-06
Quota for each project on the user dashboard


carolyncole:
Figma link - https://www.figma.com/design/rgzEYTpnGl6l68moRSVwYg/Phase-1-UI-Details?node-id=807-2898&node-type=frame&m=dev

We need a quota widget to display for each project on the user dashboard

### Current display
<img width="1111" alt="Screenshot 2024-11-06 at 11 14 24 AM" src="https://github.com/user-attachments/assets/0c99fdb9-9285-41e9-b1f5-8ebbe4997d30">

### Desired display
![383306301-ca9cd4e9-a568-4910-86eb-b9292ab40941](https://github.com/user-attachments/assets/42c08f97-2614-4662-a75c-2288e00fb541)

### Acceptance criteria
- [x] Request the data from Mediaflux
- [x] The quota and consumed storage are reflected in the UI with a bar graph as shown in Figma ([potential strategy here](https://css-tricks.com/making-charts-with-css/))
- [x] There is a test for this in the view

kelynch:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tigerdata-app/1030?planning-poker) with Zenhub @bess @carolyncole @JaymeeH @jrgriffiniii @leefaisonr

PR #1077, opened 2024-11-27
Adding project quota to the datatable

JaymeeH:
adding storage capacity and usage to the dashboard presenter
passing the session id through the welcome controller
refs #1030


JaymeeH:
Screenshot 2024-11-27 at 1 47 28 PM

JaymeeH:
Screenshot 2024-12-03 at 4 45 31 PM

JaymeeH:
Screenshot 2024-12-04 at 1 47 16 PM

app/models/project.rb
```diff
@@ -240,15 +240,14 @@ def storage_capacity_xml
   end
 
   def storage_capacity(session_id:)
-    requested_capacity = storage_capacity_xml
```
carolyncole, line 243: Can you also clean up the storage_capacity_xml function? https://github.com/pulibrary/tigerdata-app/blob/26bf465c12a856282cbd10065a0ade1b3a11d70d/app/models/project.rb#L237

app/presenters/project_dashboard_presenter.rb
```diff
@@ -40,4 +40,13 @@ def role(user)
...
     end
   end
+
+  def quota_usage(session_id:)
+    quota_usage = "#{project.storage_usage(session_id:)} out of #{project.storage_capacity(session_id:)} used"
+    quota_usage
+  end
+
+  def quota_percentage(session_id:)
+    # figure out how to calculate percentage of storage used with different and unpredictable units
```
carolyncole, line 50: I think storage is now a hash.  I wonder if there needs to be some updates to this code.  https://github.com/pulibrary/tigerdata-app/blob/main/app/models/project_metadata.rb#L149

carolyncole (changes requested):
@JaymeeH nice work.  It looks like the percentage is not getting calculated.  Was this intended? I also wonder if the fact that the size is a hash instead of a string will cause some issues. 

At the very least can you clean up the method you made unused?

carolyncole (approved):
Thanks @JaymeeH!  Nice step towards the finished product.

PR #1090, opened 2024-12-05
Ensures that the directory path is disabled for editing projects (unless the user is a system administrator)

### LINKED ISSUE #1046: The `Directory Path` field should not be writable after it is approved
In Tigerdata, the edit form displays two fields `Project Directory` which is read only, and `Directory Path` which is writable. Editing the `Directory Path` and submitting the form updates both fields. Since both fields seem to be technically displaying the same information, It would make most sense to only display one and choose whether it is read only or read/write.

Acceptance Criteria:

- [ ] Data Sponsor should view the path, but not edit it
- [ ] Data Manager should view the path, but not edit it
- [ ] Data user should view the path, but not edit it
- [ ] System Administrator should view and edit  the path

jrgriffiniii:
Resolves #1046


spec/system/project_spec.rb
```diff
@@ -5,6 +5,7 @@
 RSpec.describe "Project Page", connect_to_mediaflux: true, type: :system  do
   let(:sponsor_user) { FactoryBot.create(:project_sponsor, uid: "pul123") }
   let(:sysadmin_user) { FactoryBot.create(:sysadmin, uid: "puladmin") }
+  let(:superuser) { FactoryBot.create(:superuser, uid: "root") }
```
carolyncole, line 8: @jrgriffiniii should this line be removed since it is unused?

carolyncole (dismissed):
@jrgriffiniii It looks like you are not testing your code change anywhere, maybe you forgot to commit some changes?

kelynch:
It looks like this branch needs to be updated with the latest changes from main.  When I ran the test suite, `/spec/system/project_roles_spec.rb:209` produced an error.

kelynch (changes requested):
It looks like this branch needs to be updated with the latest changes from main. When I ran the test suite, /spec/system/project_roles_spec.rb:209 produced an error.

carolyncole (approved):
Thanks @jrgriffiniii.  Looks great!

PR #1096, opened 2024-12-09
Hide "Withdraw", "Deny", and "Custom query" buttons

### LINKED ISSUE #985: Hide non-defined functions
Hide for production

- Withdraw project
- Deny project
- Custom query

Features in development or down the roadmap should not be displayed in production for release.

Previous issue: https://github.com/pulibrary/aux/issues/8 

Acceptance Critieria:

- [x] Remove Withdraw project button
- [x] Remove Deny project button
- [x] Remove Custom query button

jrgriffiniii:
Resolves #985

### DIFF
--- a/app/views/projects/show.html.erb
+++ b/app/views/projects/show.html.erb
@@ -92,7 +92,6 @@
 <div class="details">
   <% if !current_user.eligible_sysadmin? %>
     <%= link_to "Review Contents", project_contents_path(@project.id), class: "btn btn-primary btn-sm" %>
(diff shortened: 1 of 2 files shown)


bess (changes requested):
I can see you put a lot of work into doing it this way, and it's a clever setup, but I think the original ticket was not written well. I do not think we want to create a situation where the staging and qa/training environments are showing something different from production. That's going to make it very hard to tell whether the app is behaving correctly before a production release. 

Would you please just remove the buttons? Later, we can add them back in behind a feature flipper so we can turn them on and off selectively.

bess (approved):

Issue #1147, opened 2024-12-18
SPIKE - Do we have to remove datatables


carolyncole:
Can we leave datatables in place and still get the quota and the title to float in the right place.

This may be accomplished by hiding the real data and showing another column with the visual stuff 

Or some tother way.

PR #1160, opened 2024-12-20
adding a styled column

carolyncole:
refs #1147


carolyncole:
Note to self.  When the title is not long the project details pane in not wide enough...  Deploy to staging to see this.

Issue #1210, opened 2025-01-07
Departments come in from Mediaflux as a Name not a code


carolyncole:
Departments that are imported from Mediaflux are coming in as a name not a code

Completion Criteria:

- [x] Add the [values in this sheet](https://docs.google.com/spreadsheets/d/12et7Z10I89e4BDVCCWt5nUyTVKg_f2eR/edit?gid=327298540#gid=327298540) to princeton_ansible 
- [ ] Import the department data into affiliations
- [ ] Change the import project rake task to convert the name to an affiliation code

carolyncole:
We could possibly get code from ldap for the projects we already have (reqires net_ldap gem)
```
Project.all(&:metadata_model).map(&:departments).flatten.uniq.each do |department|
  filter = Net::LDAP::Filter.eq("ou", department.gsub("Department of","") )  
  results = ldap_connection.search(filter:)     
  codes = results.map{|r| r[:puhomedepartmentnumber]}.uniq  
  puts "department #{department} codes #{ codes}"  
end  
```
So far from production this would handle all but `PNI` and `Lewis-Sigler Institute`

kelynch:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/pulibrary/tigerdata-app/1210?planning-poker) with Zenhub @bess @carolyncole @JaymeeH @jrgriffiniii @leefaisonr

kelynch:
Chuck is doing this remediation starting today.

cbentlerOIT:
The updates to the department values and the DOIs have been completed in Mediaflux.  Any new configuration will use the department value from the spreadsheet.

PR #1216, opened 2025-01-13
Allow for departments to be imported from Production

carolyncole:
Will change the name to a department code

refs #1210

### DIFF
--- a/spec/models/affiliation_spec.rb
+++ b/spec/models/affiliation_spec.rb
@@ -22,4 +22,18 @@ RSpec.describe Affiliation, type: :model do
       expect(Affiliation.count).to eq 5
     end
   end
+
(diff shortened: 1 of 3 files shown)


spec/models/affiliation_spec.rb
```diff
@@ -22,4 +22,18 @@
       expect(Affiliation.count).to eq 5
     end
   end
+
+  describe "find_find_fuzzy_by_name" do
```
JaymeeH, line 26: Is this a typo?

JaymeeH (approved):
Small typo but looks good

spec/models/affiliation_spec.rb
carolyncole: Great catch!  Thank you!

PR #1252, opened 2025-01-28
Test CI against latest mediaflux build

JaymeeH:
updating mediaflux build that circleci runs against


carolyncole (approved):
🎉 Thanks @JaymeeH

PR #1284, opened 2025-02-05
1277 emulation refinement

### LINKED ISSUE #1277: Tester trainer should only see the Create new when in emulation.  They should not see the button if they are a sponsor by default
### Expected behavior
Currently the tester trainer users see the "Create new" button by default if they also have the sponsor role

### Actual behavior
The tester trainer should not see the button by default.  Only if they are emulating the role of sponsor should they see the button.

### Acceptance criteria
- [ ] figure out how to tell if emulation is happening
- [ ] Modify eligible_to_create_new? to only allow the sponsor to see create new if the emulation is occuring

### Honeybadger link and code snippet, if applicable

```
```

JaymeeH:
closes #1277

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -94,21 +94,18 @@ class ApplicationController < ActionController::Base
       current_user.eligible_sponsor = true
       current_user.eligible_manager = false
       current_user.sysadmin = false
-      current_user.trainer = false
     end
 
     def emulate_manager
       current_user.eligible_manager = true
       current_user.eligible_sponsor = false
       current_user.sysadmin = false
-      current_user.trainer = false
     end
 
     def emulate_sysadmin
       current_user.sysadmin = true
       current_user.eligible_manager = false
       current_user.eligible_sponsor = false
-      current_user.trainer = false
     end
 
     def emulate_data_user
@@ -119,7 +116,6 @@ class ApplicationController < ActionController::Base
     end
 
     def return_to_self
-      current_user.trainer = true
       current_user.eligible_sponsor = false
       current_user.eligible_manager = false
       current_user.sysadmin = false
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -94,25 +94,28 @@ class ApplicationController < ActionController::Base
       current_user.eligible_sponsor = true
       current_user.eligible_manager = false
       current_user.sysadmin = false
+      current_user.trainer = false
     end
 
     def emulate_manager
       current_user.eligible_manager = true
       current_user.eligible_sponsor = false
       current_user.sysadmin = false
+      current_user.trainer = false
     end
 
     def emulate_sysadmin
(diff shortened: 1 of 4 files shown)


app/views/shared/_emulator.html.erb
```diff
@@ -6,11 +6,11 @@
             <form action="emulate" method="post" id="emulation-form">
                 <label for="emulation_menu" class="form-label">Current Role:</label>
                 <select id="emulation_menu" name="emulation_menu" class="form-dropdown" required>
-                    <option disabled="disabled" <%= "selected" if current_user.trainer?%>>Trainer</option>
-                    <option value="Eligible Data Sponsor" <%= "selected" if current_user.eligible_sponsor?%>>Eligible Data Sponsor</option>
-                    <option value="Eligible Data Manager" <%= "selected" if current_user.eligible_manager?%>>Eligible Data Manager</option>
-                    <option value="Eligible Data User" <%= "selected" if current_user.eligible_data_user? && !current_user.trainer?%>>Eligible Data User</option>
-                    <option value="System Administrator" <%= "selected" if current_user.sysadmin%>>System Administrator</option>
+                    <option disabled="disabled" <%= "selected" if current_user.eligible_data_user? && current_user.trainer? %>>Trainer</option>
```
carolyncole, line 9: @JaymeeH why did this change to eligible_data_user instead of trainer

app/controllers/application_controller.rb
```diff
@@ -111,12 +111,11 @@ def emulate_sysadmin
     def emulate_data_user
       current_user.eligible_sponsor = false
       current_user.eligible_manager = false
-      current_user.trainer = false
       current_user.sysadmin = false
+      current_user.trainer = false
     end
 
     def return_to_self
-      current_user.trainer = true
```
carolyncole, line 119: I'm curios why the reset on trainer was removed.  It is not making a test pass or fail.

carolyncole (commented):
Just a couple questions @JaymeeH

app/controllers/application_controller.rb
JaymeeH: The trainer here was removed becose the role never remains false since we never save the user during emulation. So returning the user, brings it back to true anyways

app/controllers/welcome_controller.rb
```diff
@@ -20,14 +20,15 @@ def index
...
+    @emulation_role = session[:emulation_role] || "Not Emulating"
   end
 
   def emulate
     return if Rails.env.production?
     return if current_user.nil? || current_user.id.nil?
 
     absolute_user = User.find(current_user.id)
-    return unless absolute_user.trainer
+    return unless absolute_user.trainer?
```
jrgriffiniii, line 31: 👍

app/views/shared/_emulator.html.erb
jrgriffiniii, line 9: 👍

jrgriffiniii (approved):

PR #1287, opened 2025-02-05
Fixes project access bug

### LINKED ISSUE #1286: URGENT: prevent project access via URL manipulation
### Expected behavior
Only those users who have specific role assignments for a given project should be able to view the project's details in the Web Portal, regardless of the method of navigating to a project page.

### Actual behavior
An early adopter (rm5876) reported to via RT (#56206):

> Hi team, wanted to flag this quickly as it’s probably a data privacy/security issue.
> 
> Logged in as myself, my dashboard shows only my own project - but I can simply navigate to any project URL other than my own, for example https://tigerdata-app.princeton.edu/projects/19 a project for which I am not an authorized user, and I can see all the files, project metadata, etc. I assume that this is not intended behavior.

This also worked for me.

### Steps to replicate

1. Log into production
2. Change the URL to https://tigerdata-app.princeton.edu/projects/19 (or any valid project number)
3. You can see the project details whether you have a role on that project or not

### Impact of this bug
This breaks the role-based access model for TigerData and raises privacy and security risks. We need a fix urgently.

### Acceptance criteria
- [ ] Only those users who have explicitly defined roles for a given project are able to load a project page (https://tigerdata-app.princeton.edu/projects/[xx])
- [ ] There is a test to prove the above

hectorcorrea:
Fixes #1286

(tests to come as a separate PR so that we can fix the bug in production quickly)

### DIFF
--- a/app/views/projects/_latest_download_link.html.erb
+++ b/app/views/projects/_latest_download_link.html.erb
@@ -1,1 +1,3 @@
-<%= link_to("Download latest report - generated #{time_ago_in_words(@latest_completed_download.completion_time)} ago", project_file_list_download_path(job_id: @latest_completed_download.job_id, target: "_blank")) %>
+<% if @latest_completed_download.completion_time %>
+  <%= link_to("Download latest report - generated #{time_ago_in_words(@latest_completed_download.completion_time)} ago", project_file_list_download_path(job_id: @latest_completed_download.job_id, target: "_blank")) %>
+<% end %>
--- a/data/user_registration_list_development.csv
+++ b/data/user_registration_list_development.csv
@@ -5,7 +5,7 @@ eparham,,Beth,Holtz,Beth Holtz,TRUE,TRUE,,TRUE,,2023-12-05,Matt Chandler,TigerDa
 cbentler,,Chuck,Bentler,Chuck Bentler,TRUE,TRUE,,TRUE,,2023-12-05,Matt Chandler,TigerData Staff
 knight,,Robert,Knight,Robert Knight,TRUE,TRUE,,TRUE,,2023-12-05,Matt Chandler,TigerData Staff
 kl37,,Kate,Lynch,Kate Lynch,TRUE,TRUE,TRUE,,,2023-12-05,Matt Chandler,TigerData Staff
-hc8719,,Hector,Correa,Hector Correa,,,TRUE,,,2023-12-05,Matt Chandler,TigerData Staff
+hc8719,,Hector,Correa,Hector Correa,,,FALSE,,,2023-12-05,Matt Chandler,TigerData Staff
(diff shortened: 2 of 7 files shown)


carolyncole (approved):
Thanks @hectorcorrea.  Looking forward to the tests too!

hectorcorrea:
Several tests are failing I think because now the system is enforcing who has access to the projects (whereas we were not before.) 

We should fix these tests or at least make sure the PR is really not breaking the project controller before pushing this to production.

app/views/projects/_latest_download_link.html.erb
```diff
@@ -1 +1,3 @@
-<%= link_to("Download latest report - generated #{time_ago_in_words(@latest_completed_download.completion_time)} ago", project_file_list_download_path(job_id: @latest_completed_download.job_id, target: "_blank")) %>
\ No newline at end of file
+<% if @latest_completed_download.completion_time %>
```
hectorcorrea, line 1: I seem to have one "completed" job without a completion time on my local environment. This handles that weird case. Not sure how that record got there but I just added the check so that it does not blow up on the user if it ever happens in prod.

spec/controllers/projects_controller_spec.rb
```diff
@@ -16,7 +16,9 @@
     end
 
     context "a signed in user" do
-      let(:user) { FactoryBot.create :user }
+      # TODO: Switch for a user with less access than system admin
```
hectorcorrea, line 19: I force the user to be a sysadmin so that the test works as it did before we had a check for access to the project. We should eventually update these tests to use a more restrictive user (but with access to the project), hence the TODO.

data/user_registration_list_development.csv
```diff
@@ -5,7 +5,7 @@ eparham,,Beth,Holtz,Beth Holtz,TRUE,TRUE,,TRUE,,2023-12-05,Matt Chandler,TigerDa
 cbentler,,Chuck,Bentler,Chuck Bentler,TRUE,TRUE,,TRUE,,2023-12-05,Matt Chandler,TigerData Staff
 knight,,Robert,Knight,Robert Knight,TRUE,TRUE,,TRUE,,2023-12-05,Matt Chandler,TigerData Staff
 kl37,,Kate,Lynch,Kate Lynch,TRUE,TRUE,TRUE,,,2023-12-05,Matt Chandler,TigerData Staff
-hc8719,,Hector,Correa,Hector Correa,,,TRUE,,,2023-12-05,Matt Chandler,TigerData Staff
+hc8719,,Hector,Correa,Hector Correa,,,FALSE,,,2023-12-05,Matt Chandler,TigerData Staff
```
hectorcorrea, line 8: Made myself a non-admin in development to make sure I can test the logic. Will probably undo this at a later time.

hectorcorrea:
Closed in favor of https://github.com/pulibrary/tigerdata-app/pull/1290

Issue #1272, opened 2025-01-31
Allow Administrator to import projects from Mediaflux


carolyncole:
### User story

As an Administrator I need the ability to import additional production projects so I can keep the front end up to date with what is currently in Mediaflux.

1. the will be a script in mediaflux that creates CSV formatted data, keeping the format the same as the current rake task
    `asset,path,creatorDomain,creatorUser,createdOn,quota,store,projectDirectory,title,description,dataSponsor,dataManager,dataUser,department,projectID`
1. The administrator will click a button "Import Mediaflux Projects" on the Administration dashboard, which will run the script and then import the csv data into the server4. If a DOI exists nothing gets created or updated
1. if a typo gets put into the DOI an email get sent to the Dev team to update the DOI
### Acceptance criteria

- [x] The mediaflux script is created (RC) `tigerdata.project.export.generate`
- [x] The script is installed in the developer docker container
- [x] The script is installed on td-meta1 & td-metalib1 (RC)
- [ ] The script is installed on production mediaflux  (RC) 
- [x] The current Rake task code is move into a service object that can be called from the UI or the rake task
- [x] There is a mediaflux command to run the script and parse the results to return the CSV data
- [x] There is a UI button "Import Mediaflux Projects" on the administrator dashboard that cause the script to be run and the service object to be called
- [x] We test the button with Research Computing to make sure it is behaving as desired in td-meta1 & td-metalib1

carolyncole:
testing on qa and staging shows a cookie overflow when many projects have already been imported.  Need to create an html page to show the results instead on a flash message.

PR #1289, opened 2025-02-06
Add a mediaflux command for calling the tigerdata project list command

carolyncole:
Also update circle-ci to use the new mediaflux docker container version with the command

refs #1272

### DIFF
--- a/app/models/mediaflux/project_report.rb
+++ b/app/models/mediaflux/project_report.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+module Mediaflux
+  class ProjectReport < Request
+    # Specifies the Mediaflux service to use when pulling a report of projects in mediaflux
+    # @return [String]
+    def self.service
+      "tigerdata.project.export.generate"
+    end
+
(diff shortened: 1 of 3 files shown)


app/models/mediaflux/project_report.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+module Mediaflux
+  class ProjectReport < Request
+    # Specifies the Mediaflux service to use when creating assets
```
hectorcorrea, line 4: The comment is wrong, right? Could we remove it?

hectorcorrea (approved):

hectorcorrea:
@carolyncole Look great. I left a nitpicking note if you want to take care of that, but it's approved so feel free to merge it as-is.

app/models/mediaflux/project_report.rb
carolyncole: Thanks for the catch @hectorcorrea!  I went ahead and updated the comment.  Copy 🍝 error

PR #1292, opened 2025-02-06
Document how to remove a machine from the load balancer

hectorcorrea:
Closes https://github.com/pulibrary/tigerdata-app/issues/1288


carolyncole (approved):
Thanks @hectorcorrea

PR #1299, opened 2025-02-10
Separating the welcome and the dashboard

carolyncole:


hectorcorrea (approved):
Thank you @carolyncole

PR #1303, opened 2025-02-11
Fix CSRF issue with tabs in the dashboard

carolyncole:
Also remove dash-classic as it is no longer in use


app/javascript/entrypoints/dashboardTabs.js
```diff
@@ -91,27 +91,18 @@ export function dashStyle(railsSession) {
...
+    const tokenElements = document.getElementsByName('csrf-token');
+    let headers = {};
+    if (tokenElements[0]) headers = { 'X-CSRF-Token': tokenElements[0].content };
+
     element.preventDefault();
     $.ajax({
       type: 'POST',
       url: '/dash_project',
       data: { dashtab: 'project' },
+      headers,
```
hectorcorrea, line 105: This syntax is news to me but it works!

hectorcorrea (approved):

app/javascript/entrypoints/dashboardTabs.js
carolyncole: Yes, new to me too.  eslint changed it.

PR #1305, opened 2025-02-12
Allow flash message to be collapsible and not overlap with other UI elements

carolyncole:
Before the Update
Screenshot 2025-02-12 at 10 09 07 AM

Screenshot 2025-02-12 at 10 18 30 AM


JaymeeH (approved):
Thanks!

Issue #1274, opened 2025-01-31
Why do file inventory jobs take so long for some users


hectorcorrea:
While fixing issue #1263 we noticed that the file inventory job takes a very long time when Matt (`user_id 186`) runs it in QA. The file inventory for `project_id 44` (which has 189,000 files) takes 65 minutes when Matt runs it, but as shown below it only takes 5 minutes when Hector `user_id 192` runs it.

Below is what the `user_request` table shows in QA.

```
#<FileInventoryRequest:0x00007ff447b7da20
  id: 16,
  user_id: 192,
  project_id: 44,
  job_id: "fa74522e-c720-4ed5-b30b-741bdeb3b804",
  completion_time: Fri, 31 Jan 2025 17:15:35.378825000 UTC +00:00,
  state: "completed",
  type: "FileInventoryRequest",
  request_details: {"file_size"=>22090419, "output_file"=>"/mnt/nfs/tigerdata/fa74522e-c720-4ed5-b30b-741bdeb3b804.csv", "project_title"=>"Test 06-12-24"},
  created_at: Fri, 31 Jan 2025 17:10:12.237105000 UTC +00:00,
  updated_at: Fri, 31 Jan 2025 17:15:35.382571000 UTC +00:00>,
 #<FileInventoryRequest:0x00007ff447b7d660
  id: 17,
  user_id: 186,
  project_id: 44,
  job_id: "fdb79db4-e6b6-41a3-842a-b63afa619163",
  completion_time: Fri, 31 Jan 2025 20:05:34.709517000 UTC +00:00,
  state: "completed",
  type: "FileInventoryRequest",
  request_details: {"file_size"=>22090419, "output_file"=>"/mnt/nfs/tigerdata/fdb79db4-e6b6-41a3-842a-b63afa619163.csv", "project_title"=>"Test 06-12-24"},
  created_at: Fri, 31 Jan 2025 18:59:57.304534000 UTC +00:00,
  updated_at: Fri, 31 Jan 2025 20:05:34.719367000 UTC +00:00>]
```

We think this could be related to the fact that Hector is an admin on MediaFlux QA but Matt isn't.

hectorcorrea:
In the QA MediaFlux instance my permissions are very different from Matt's permissions:

```
> user.authority.describe :domain princeton :user hc8719
    :user -domain "princeton" -user "hc8719"
        :role -type "domain" "princeton"
        :role -type "role" "user"
        :role -type "role" "standard-user"
[...]

PR #1335, opened 2025-03-04
Rake task to run a FileInventoryJob from the command line

hectorcorrea:
Added a Rake task that kicks off the FileInventoryJob for a given project. It assigns the job to the user indicated in the Rake task but allows you to specify the user credentials (domain,user,password) that will be used to connect to MediaFlux. This is to allows us to test with different accounts outside of the CAS authentication.

```
bundle exec rake:file_inventory:run[65,hc8719]
```

Where `65` is the `project_id` and `hc8719` is the `netid` of the user that the job will be assigned to. In the example above the job will show under my own Dashboard.

Notice that the MediaFlux credentials are prompted on the command line and the password is not written to the Rails log.

Part of #1274


carolyncole (approved):
Looks good @hectorcorrea

PR #1346, opened 2025-03-20
Cleanup daily run

JaymeeH:
ref #1345


carolyncole (approved):
Thanks @JaymeeH .  I don't really understand what is going on here, but it looks ok to me

Issue #1359, opened 2025-04-03
Update registration list 2025-04-03


matthewjchandler:
Please update the web app permissions following the newly updated [Preliminary Registration List](https://docs.google.com/spreadsheets/d/169lfRTOSe6H66Iu2DK5g-QzqiVsfz5lHFHMaGmwNT7Y/edit?gid=0#gid=0)

PR #1360, opened 2025-04-04
 updating the user registration list

JaymeeH:
refs #1359


carolyncole (approved):
Approving even though CI is failing because we want these users in prod

Issue #1426, opened 2025-04-24
User reports missing project on dashboard


matthewjchandler:
### Expected behavior
Users should see every project on which they have a role in the frontend.

### Actual behavior
The data manager (NetID: yp1065) for the project "BRD4 transcriptional condensates on chromatin" (DOI:10.60803/wszm-k791) cannot see that project on their dashboard (screenshot provided in [RT #57940](https://cses.princeton.edu/tickets/Ticket/Display.html?id=57940)).

### Impact of this bug
The data manager is not able to monitor their storage usage directly, so they have to ask the backend team for help.

### Acceptance criteria
- [ ] "BRD4 transcriptional condensates on chromatin" (DOI:10.60803/wszm-k791) exists in the frontend
- [ ] yp1065 is assigned as the data manager on the above project in the frontend
- [ ] Nothing is blocking yp1065 from seeing all of their projects on the dashboard
- [ ] If this change touches any of the files referenced in `.rubocop_todo.yml` create a new ticket to address the rubocop  todo.

JaymeeH:
@kelynch  I don't have permission to view this RT ticket ( [RT #57940](https://cses.princeton.edu/tickets/Ticket/Display.html?id=57940) ) How should I go about requesting access?

PR #1438, opened 2025-04-29
Updating user registration list

JaymeeH:
ref #1426


matthewjchandler:
> @matthewjchandler Looks like this user was not listed as an eligible_manager in the user registration list. I updated it and I can release it with your approval

Thanks for catching that @JaymeeH. I approve your update here in the codebase. I also updated the permissions for yp1065 in https://docs.google.com/spreadsheets/d/169lfRTOSe6H66Iu2DK5g-QzqiVsfz5lHFHMaGmwNT7Y/edit?gid=0#gid=0

PR #1463, opened 2025-05-09
Script to create v0.8 schema in Mediaflux

hectorcorrea:
Starting to create the script to create a valid v0.8 schema in MediaFlux.


carolyncole (approved):
Thanks @hectorcorrea

PR #1469, opened 2025-05-12
Adjusted CSS to utilize variables for colors and typography across all stylesheets

### LINKED ISSUE #1468: Use CSS variables uniformly
In some places, we are using CSS variables defined in the `variables` and `typography` stylesheets, and in others we are hard-coding these values.  Let's standardize and use the variables where we are otherwise hard-coding their values.

kelynch:
Closes #1468


leefaisonr (approved):
Thanks @kelynch

PR #1501, opened 2025-05-22
Adding the storage options to the new request wizard

carolyncole:
Realized we needed a bit more storage in the request to allow for easily back and forth to the wizard


hectorcorrea (approved):
Looks great!

Issue #1498, opened 2025-05-21
User validation with LDAP


kelynch:
Related to https://github.com/pulibrary/tigerdata-app/issues/1479 

When a net ID is entered into Skeletor, it is validated against LDAP to ensure that it is a valid net ID.  If the net ID is valid but the user does not yet exist, we create the user.  We also want to grab the user's given name and family name and populate those fields on the user object.  

This would replace the current workflow of adding users via spreadsheet/rake task as the only way to get users in, but we should not retire the rake task yet.

## Acceptance criteria
- [ ] Write an LDAP lookup/validation function
- [ ] The fields where net IDs are entered in Skeletor have a validation against this function applied
- [ ] If valid, do a user lookup in our application and create the user if they do not exist
- [ ] Populate the user's family name and given name fields with what is returned from LDAP

kelynch:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/gh/pulibrary/tigerdata-app/1498?planning-poker) with Zenhub @bess @carolyncole @hectorcorrea @JaymeeH @jrgriffiniii @leefaisonr

carolyncole:
Roles do get removed if the CSV file is reloaded. 

I think the right thing to do for now is to Add roles to the user when the request gets approved.  What do you think @matthewjchandler?  Can we turn off the reloading off the CSV file on deployment and add roles to the users as requests get approved by administrators?

matthewjchandler:
@carolyncole yes, we want role eligibility to be added when a request is approved in the application. We don't want the absence of an eligibility in the CSV to overwrite after approval takes place in the application.

But if we turn off the reloading of the CSV now, what would be the potential impact for the live usage of the system? Are we ready yet to make the hard switch?

PR #1520, opened 2025-06-03
Allow the princeton users to be loaded from ldap

carolyncole:
This occurs each morning via a cron job.  Users will  get updated if thier display name is blank or created if the uid is missing

refs #1498


spec/services/princeton_users_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe PrincetonUsers, type: :model do
+  describe "#create_users_from_ldap" do
+    # this test can not be run on circle (tagged :no_ci to not run on circle ci)
```
bess, line 7: Looks like we've got the start of our integration test suite here. Thank you!

bess (approved):
🚀 This is great progress, thank you!

Issue #1482, opened 2025-05-14
A DOI is minted when the project request is approved


kelynch:
Builds on https://github.com/pulibrary/tigerdata-app/issues/1483

There is a separate DOI prefix and account for TigerData; details in LastPass for login.  TigerData prefix: 10.60803

## Acceptance criteria 

- [ ] Configure the application to use the DOI sandbox for everything but production, and to only use the TigerData credentials in production
- [ ] After a request is approved, during the data conversion stage, mint a DOI as a draft
- [ ] Populate the project ID field with the DOI
- [ ] Allow the project to be saved

## Notes
Notice that unlike PDC Describe, we won't be making the DOI public at this time in TigerData.

kelynch:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/gh/pulibrary/tigerdata-app/1482?planning-poker) with Zenhub @bess @carolyncole @hectorcorrea @JaymeeH @jrgriffiniii @leefaisonr

PR #1534, opened 2025-06-04
Add an additional test for the custom quota conversion

carolyncole:
Also add some comments for where #1482 should be tackled when the ticket is picked back up


leefaisonr (approved):
Thanks @carolyncole

PR #1540, opened 2025-06-05
Fixed Storage size tab interactions

### LINKED ISSUE #1524: Project Storage Size does not get  tab focus
Need to add the ability to focus on the storage sizes when tabbing

Set focus to true and tabindex to 0 like https://github.com/pulibrary/tigerdata-app/blob/main/app/javascript/entrypoints/departmentAutocomplete.js#L28-L29

carolyncole:
Also allow the custom size and unit to be saved by allowing the parameters

fixes #1524


JaymeeH (approved):
Thanks! @carolyncole

PR #1581, opened 2025-06-27
Remove schema create and update functionality from tigerdata-app

JaymeeH:
Remove the code from tigerdata-app that updates the schema because we should no longer be updating schemas with ruby code; we want that definition managed via a tcl script in the mediaflux config (remove it from the test setup too)

ref [TD-Config-114](https://github.com/PrincetonUniversityLibrary/tigerdata-config/issues/114)

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -71,19 +71,6 @@ namespace :mailcatcher do
   end
 end
 
-# namespace :schema do
-#   desc "Load the current schema"
-#   task :load do
-#     on roles(:schema) do
-#       within release_path do
-#         execute("cd #{release_path} && bundle exec rake schema:create")
-#       end
-#     end
-#   end
-# end
-
-# after "deploy:published", "schema:load"
-
 before "deploy:reverted", "npm:install"
 
 namespace :sidekiq do


app/models/mediaflux/schema_fields_create_request.rb
```diff
@@ -1,64 +1,65 @@
 # frozen_string_literal: true
```
hectorcorrea, line 1: @JaymeeH could we delete this file instead of commenting all lines of it

app/models/tigerdata_schema.rb
```diff
@@ -67,107 +67,107 @@ def create_aterm_doc_script(filename: Rails.root.join("docs", "schema_script.txt
   def self.project_schema_fields
     # WARNING: Do not use `id` as field name, MediaFlux uses specific rules for an `id` field.
     project_directory = { name: "ProjectDirectory", type: "string", index: true, "min-occurs" => 1, "max-occurs" => 1, label: "Project Directory", description: "The locally unique name for the project's top-level directory",
-                          instructions: "The locally unique name for the project's top-level directory, as shown in a file path. Data Sponsors may suggest a project directory name that is meaningful to them, subject to system administrator approval." 
+                          instructions: "The locally unique name for the project's top-level directory, as shown in a file path. Data Sponsors may suggest a project directory name that is meaningful to them, subject to system administrator approval."
```
hectorcorrea, line 70: @JaymeeH is there any reason we are not removing this method (`def self.project_schema_fields`) too?

spec/controllers/projects_controller_spec.rb
```diff
@@ -61,25 +61,12 @@
           "          <ProjectDirectory>#{project.project_directory}</ProjectDirectory>\n" \
           "          <Title>#{project.metadata[:title]}</Title>\n" \
           "          <Description>#{project.metadata[:description]}</Description>\n" \
-          "          <Status>#{project.metadata[:status]}</Status>\n" \
```
hectorcorrea, line 64: @JaymeeH good catch, I forgot the "status" does not exist in the real schema

hectorcorrea:
@JaymeeH awesome work here.

app/models/tigerdata_schema.rb
JaymeeH: Should we just delete the whole file?

hectorcorrea: I am not sure. I think we will end up deleting it all but I saw a few other functions defined here that I am not sure.

config/deploy.rb
```diff
@@ -71,18 +71,18 @@
...
-  desc "Load the current schema"
-  task :load do
-    on roles(:schema) do
-      within release_path do
-        execute("cd #{release_path} && bundle exec rake schema:create")
-      end
-    end
-  end
-end
+# namespace :schema do
```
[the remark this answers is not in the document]
JaymeeH: Deleted 👍🏿

app/models/tigerdata_schema.rb
JaymeeH: Yea we need to be able to reference _a_ schema when it comes to project validation, we should talk about how we wanna refactor that

Issue #1625, opened 2025-07-22
Review tests that use td-test-001


hectorcorrea:
We have a few tests that still reference `td-test-001` and I don't think we need them. I believe they should either be removed or updated to point to `princeton/tigerdata/rspec`

```
1) Mediaflux::NamespaceCreateRequest#resolve disconnects the session
     Failure/Error: raise(StandardError, "call to service 'asset.namespace.hard.destroy' failed: The namespace #{namespace} does not exist or is not accessible")
     
     StandardError:
       call to service 'asset.namespace.hard.destroy' failed: The namespace td-test-001 does not exist or is not accessible
     # ./app/models/mediaflux/namespace_destroy_request.rb:26:in `destroy'
     # ./spec/models/mediaflux/namespace_create_request_spec.rb:13:in `block (3 levels) in <top (required)>'
     # /Users/correah/.gem/ruby/3.3.0/gems/webmock-3.23.1/lib/webmock/rspec.rb:39:in `block (2 levels) in <top (required)>'
     # /Users/correah/.gem/ruby/3.3.0/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:124:in `block in run'
     # /Users/correah/.gem/ruby/3.3.0/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:110:in `run'
     # /Users/correah/.gem/ruby/3.3.0/gems/rspec-retry-0.6.2/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'
     # /Users/correah/.gem/ruby/3.3.0/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:37:in `block (2 levels) in setup'

```

PR #1629, opened 2025-07-22
Updated test not to reference td-test-001

hectorcorrea:
Updated one of the tests not to reference `td-test-001` since we are moving away from it now that everything starts at `/princeton`

Moves #1625 forward.


hectorcorrea:
@JaymeeH that's a different test than the one I fixed on this PR. I fixed `namespace_create_request_spec.rb` and the one failing is `namespace_list_request_spec.rb` so I would say let's approve this. 

I'll take a look at the failing one separately.

PR #1633, opened 2025-07-23
Adjust the test to create test assets within the new root

hectorcorrea:
Moves https://github.com/pulibrary/tigerdata-app/issues/1625 forward


hectorcorrea:
@JaymeeH this is ready for review, notice that the failing tests is not the one that I touched on this PR

JaymeeH (approved):
Thanks @hectorcorrea

PR #1635, opened 2025-07-23
Fixes another test to use the new default namespaces

hectorcorrea:
Keep moving https://github.com/pulibrary/tigerdata-app/issues/1625 forward


hectorcorrea:
Fixes the test that shows us as broken in https://github.com/pulibrary/tigerdata-app/pull/1634

PR #1640, opened 2025-07-24
Refactored code that produces the XML representation

hectorcorrea:
Refactored the XML produced when visiting `projects/:id.xml` to not fetch from Mediaflux (since we only need the Ruby data) and the code when visiting `projects/:id/:id-mf.xml` to just fetch the data from Mediaflux (there was a call to create in Mediaflux buried in there.

Closes https://github.com/pulibrary/tigerdata-app/issues/1607

### DIFF
--- a/spec/controllers/projects_controller_spec.rb
+++ b/spec/controllers/projects_controller_spec.rb
@@ -46,41 +46,16 @@ RSpec.describe ProjectsController, type: ["controller", "feature"] do
         end
 
         it "renders the project metadata as xml" do
-          data_sponsor = project.metadata_model.data_sponsor
-          project = FactoryBot.create :project, project_id: "abc-123", data_sponsor: data_sponsor
+          project.approve!(current_user: sponsor_and_data_manager)
           get :details, params: { id: project.id, format: :xml }
           expect(response.content_type).to eq("application/xml; charset=utf-8")
-          expect(response.body).to eq(
-          "<?xml version=\"1.0\"?>\n" \
-          "<request xmlns:tigerdata=\"http://tigerdata.princeton.edu\">\n" \
-          "  <service name=\"asset.create\">\n" \
-          "    <args>\n" \
-          "      <name>#{project.project_directory_short}</name>\n" \
-          "      <namespace>#{project.project_directory}NS</namespace>\n" \
-          "      <meta>\n" \
-          "        <tigerdata:project>\n" \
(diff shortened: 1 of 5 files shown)


spec/controllers/projects_controller_spec.rb
```diff
@@ -46,41 +46,29 @@
...
+          xml_doc = Nokogiri::XML(response.body)
+          expect(xml_doc.xpath("/resource/title").text).to eq project.title
+          expect(xml_doc.xpath("/resource/projectDirectory/projectDirectoryPath").text).to eq project.project_directory
+          expect(xml_doc.xpath("/resource/description").text).to eq project.metadata[:description]
+          expect(xml_doc.xpath("/resource/dataSponsor/@userID").text).to eq project.metadata[:data_sponsor]
+          expect(xml_doc.xpath("/resource/dataManager/@userID").text).to eq project.metadata[:data_manager]
+          expect(xml_doc.xpath("/resource/projectID").text).to eq project.metadata[:project_id]
+        end
+
+        it "renders the Mediaflux metadata as xml" do
```
hectorcorrea, line 61: This is the test for the Mediaflux version of the data.

spec/controllers/projects_controller_spec.rb
```diff
@@ -45,42 +45,30 @@
...
+          project.approve!(current_user: sponsor_and_data_manager)
+          get :show_mediaflux, params: { id: project.id, format: :xml }
+          expect(response.content_type).to eq("application/xml; charset=utf-8")
+          xml_doc = Nokogiri::XML(response.body)
+          expect(xml_doc.xpath("/meta//Title").text).to eq project.title
+          expect(xml_doc.xpath("/meta//ProjectDirectory").text).to eq project.project_directory
+          expect(xml_doc.xpath("/meta//Description").text).to eq project.metadata[:description]
+          expect(xml_doc.xpath("/meta//DataSponsor").text).to eq project.metadata[:data_sponsor]
+          expect(xml_doc.xpath("/meta//DataManager").text).to eq project.metadata[:data_manager]
+          expect(xml_doc.xpath("/meta//ProjectID").text).to eq project.metadata[:project_id]
```
hectorcorrea, line 71: This new tests is the one that checks for the values coming from Mediaflux (notice the `meta//Title` xpath)

carolyncole (approved):
Nice work @hectorcorrea.  Thanks for deleting the old unused code

PR #1644, opened 2025-07-25
Updated test to use new location of data in Mediaflux

hectorcorrea:
Makes sure the test looks for data in the new location that we use in the Docker container that mimics production


carolyncole (approved):
@hectorcorrea this change is fine.  There are still a number of locations that refernce api_root https://github.com/search?q=repo%3Apulibrary%2Ftigerdata-app%20api_root_collection_namespace&type=code

hectorcorrea:
@carolyncole yes, indeed there is a lot more to clean up. More (small) PRs are on the way 🤣

PR #1650, opened 2025-07-28
Make sure our accounts are administrators on our local docker

hectorcorrea:
Makes sure our local accounts are system administrators on Mediaflux in our local Docker.


lib/tasks/servers.rake
```diff
@@ -20,11 +20,27 @@ namespace :servers do
...
   end
+
+  desc "Make sure the developer accounts are administrators"
+  task dev_admin_users: :environment do
+    login = Mediaflux::LogonRequest.new
+    login.resolve
+
+    type = "role"
+    name = "pu-lib:developer"
+    role = "system-administrator"
```
hectorcorrea, line 38: This is the magic sauce

carolyncole (commented):
@Hector The code looks good.  My only concern is if we are planning to delete this code once tigerdata-config does the right thing?  If yes, maybe a TODO.  If no a test would be nice.

PR #1657, opened 2025-07-28
Add data users to a project after it is created in mediaflux

### LINKED ISSUE #1632: Save Data Users in Mediaflux
The approval process now creates the project in Mediaflux via the `tigerdata.project.create` service. But this service does not accept Data Users (neither read only not read-write users), we need to make a separate call to add the data users to the project once it has been created in Mediaflux.

Reference https://github.com/PrincetonUniversityLibrary/tigerdata-config/blob/main/roles/mflux/files/services/service_scripts/tigerdata.project.user.add.tcl 

## acceptance criteria

- [x] Call `tigerdata.project.user.add` after `tigerdata.project.create` is successfully called
- [x] Wrap this transaction for success or failure using https://dry-rb.org/gems/dry-operation/1.0/
  - [x] If the project creation fails, bail out and report the error
  - [x] If the project is created but the user add fails, bail out and report the error 
- [x] There is a test for this

JaymeeH:
closes #1632


app/models/request.rb
```diff
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+# Explicitly require the file from your model.
+require_relative "../operations/project_create"
```
bess, line 3: I'd love to avoid one-off require statements if we can. These cause confusing bugs down the road. I would think that the operations folder, by virtue of being inside of the app folder, would just be loaded? Or maybe there is a load_path that needs adjusting?

bess (changes requested):
Overall this looks great. I love your use of inline comments and documentation for mediaflux interactions particularly. 

I would like to avoid the use of one-off require statements, however. Can you please take a pass at finding another way to load that newly created operations folder?

bess (approved):

PR #1666, opened 2025-07-30
Refactor project creation in Mediaflux

hectorcorrea:
Consolidated all the logic for project creation in Mediaflux to happen via the `Project.approve!` method. 

We had too many different ways of creating a project in Mediaflux (`Project.approve!`, `Project.save_in_mediaflux`, `ProjectMediaflux.save`, and `ProjectMediaflux.create!`). In the end all of them funneled the logic through the `Project.approve!` method but it was confusing to know which method to call and with all the refactoring to the code there seem to be no rhyme or reason for each of them anymore. 

With this PR there is now only one call that creates a project in Mediaflux: `Project.approve!` I suspect this code will be eventually converted into an operation (as part of Jaymee's PR https://github.com/pulibrary/tigerdata-app/pull/1657) but the current PR just consolidates the code. 

We don't support updating the project in Mediaflux yet (and we never really did). Once we have a Mediaflux service to update projects in Mediaflux we can implement that but that feature is not scheduled until we deliver SpaceGhost.


carolyncole (approved):
Thanks @hectorcorrea

PR #1671, opened 2025-08-01
Fetches schema definition from Mediaflux

### LINKED ISSUE #1627: Fetch schema definition from Mediaflux
Our current TigerdataSchema defines the schema in Ruby (see https://github.com/pulibrary/tigerdata-app/blob/main/app/models/tigerdata_schema.rb) but now that the schema comes pre-defined in the Docker image we should fetch it from Mediaflux so that we don't have this information in two places.

We currently only use this information to calculate the list of fields that are required (`min-occurs = 1`)

Acceptance Requirement:

- [ ] Get the schema from mediaflux
- [ ] parse the fetched code to set a list of required fields
- [ ] remove our TigerdataSchema from codebase
- [ ] remove schema create in models mediaflux

hectorcorrea:
Fetch schema definition directly from Mediaflux instead of redefining it in the Ruby codebase. At this time we only use this definition to detect that list of fields that are required (i.e. `min-occurs > 1` in Mediaflux lingo.)

Closes #1627


app/models/tigerdata_schema.rb
```diff
@@ -1,162 +1,22 @@
...
-      fields: project_schema_fields,
-      session_token: @session_id
-    )
-    fields_request.resolve
-    if fields_request.error?
-      raise "Could not create or update schema #{fields_request.response_error}"
+  def fields
+    @fields ||= begin
+      logon_request = Mediaflux::LogonRequest.new
+      schema_request = Mediaflux::SchemaFetchRequest.new(session_token: logon_request.session_token, namespace: @namespace, type: @type)
```
hectorcorrea, line 13: This fetches the field definition from Mediaflux rather than relying on our hard-coded version in Ruby (which is being deleted below, see lines 42-150)

bess (approved):

PR #1708, opened 2025-08-08
Improve exception handling for project creation

### LINKED ISSUE #1706: Exception handling on project creation
The code that handles exception in the Request model seems to be ignored when the exception is NOT a `ProjectCreate::ProjectCreateError` as it happened when there were no defaults for the project values in the configuration (see PR #1705). 

Perhaps we should catch Standard Exception here rather than `ProjectCreate::ProjectCreateError`

We should also log the error in the Rails log and in HoneyBadger.

```
def approve(approver)
    create_project_operation = ProjectCreate.new
    result = create_project_operation.call(request: self, approver: approver)
    result = result.flatten while result.class != Project
    result
  rescue ProjectCreate::ProjectCreateError => ex
    # Save the error within the Request object
    self.error_message = { message: ex.message }
    save!
    # ..and get rid of the Rails project
    project.destroy!
    raise "Error approving request #{id}"
  end
```

We found this while troubleshooting https://github.com/pulibrary/tigerdata-app/issues/1691

hectorcorrea:
This PR improves the exception handling for project creation (the code now logs _any_ exception) and removes the duplicate project creation code.

In order to keep this PR small I left a stub `project.approve!()` method that will eventually needs to be removed (see ticket https://github.com/pulibrary/tigerdata-app/issues/1707)

Closes #1706


app/models/project.rb
```diff
@@ -35,37 +33,15 @@ def create!(initial_metadata:, user:)
     end
   end
 
+  # TODO: Remove this method https://github.com/pulibrary/tigerdata-app/issues/1707 has been completed
   def approve!(current_user:)
-    request = Mediaflux::ProjectCreateServiceRequest.new(session_token: current_user.mediaflux_session, project: self)
```
hectorcorrea, line 39: This code was duplicated with `ProjectCreate.new`

app/models/project.rb
```diff
@@ -35,37 +33,15 @@ def create!(initial_metadata:, user:)
...
-    # - another with debug information from the create project service
-    ProvenanceEvent.generate_approval_events(project: self, user: current_user, debug_output: debug_output)
-
+    # This code is duplicated with Request.approve() and it should
+    # be removed. We keep it for now since we have way too many tests
+    # wired to it already. The goal is that projects won't be approved,
+    # instead Request are approved (and that process creates the project)
+    create_project_operation = ProjectCreate.new
+    result = create_project_operation.call(request: nil, approver: current_user, project: self)
+    result = result.flatten while result.class != Project
```
carolyncole, line 44: @hectorcorrea what does this do?  I would have expected a check of the result class ala https://github.com/pulibrary/orcid_princeton_hanami/blob/7ac21bf6b91d05af839b392751c0966b1aacae7b/app/actions/admin/orcid_report.rb#L20

carolyncole (changes requested):
I'm unsure about how you are checking the result of the operation.  Other than that this looks ok.

app/models/project.rb
hectorcorrea: Good question. I copied this from the original code that uses the operation https://github.com/pulibrary/tigerdata-app/blob/main/app/models/request.rb#L69-L73 so I thought it was a standard thing.

hectorcorrea: @JaymeeH since you did this on the original code do you have any ideas on this?

JaymeeH: The flatten operation unwraps the layers of Success objects around the the Project object. When I wrote the original code, I expected their to only be one Succes object wrapping the returned project but it ended up being multiple layers. If there is a better way to grab that value, we can rework it.

PR #1760, opened 2025-08-20
Adds step to activate a project

### LINKED ISSUE #1726: Make project Active
After a project has been approved we should query Mediaflux to make sure some basic metadata was saved (e.g. the title) and if that is the case then we mark the project as Active.

hectorcorrea:
Closes #1726


carolyncole (approved):
Thanks @hectorcorrea

PR #1785, opened 2025-08-26
Data users fetched via AJAX 

### LINKED ISSUE #1763: Address performance issues in Data Users form element
As noted in https://github.com/pulibrary/tigerdata-app/issues/1748, the Data Users dropdowns are very laggy in staging and QA using the HTTP method for creating many HTML elements before loading the form element.

We should instead implement an AJAX method in the controller to look up users in the database in real time based on what they are typing into the text box.

We refresh our list of users from LDAP daily at 1:00AM UTC.

## Acceptance criteria

- [ ] The new lookup method is implemented in the controller and on the webform.
- [ ] When a user enters a data user, they see them on the final review screen.
- [ ] Validate that the performance issue is addressed on staging.

hectorcorrea:
Closes #1763


app/services/princeton_users.rb
```diff
@@ -12,6 +12,11 @@ def user_list
       end
     end
 
+    def user_list_query(query)
+      return [] if query == ""
+      user_list.select { |user| user[:uid].downcase.include?(query) || user[:name]&.downcase&.include?(query) }
```
carolyncole, line 17: Original search allows for tokens.  @hector will do this in a separate PR

carolyncole (approved):
Thanks @hectorcorrea!  So much faster!

PR #1804, opened 2025-08-29
Make sure we don't leave orphan projects behind

### LINKED ISSUE #1800: Duplicate project created when approving a project
When approving a project Andrew session time out and the system threw an error indicating the error. Andrew logged back in and approved the project again and it worked. 

However the system left the initial project created in the database, see screenshot below. The project that is approved (but not active) does not exist in Mediaflux, only in our Rails database. The approved project is the one that has a matching asset in Mediaflux.

Image

See the Honeybadger exception for the session timeout here: https://app.honeybadger.io/projects/113559/faults/122678715

One good thing is that the Honeybadger error message clearly indicated that the problem was a session timeout (see https://github.com/pulibrary/tigerdata-app/issues/1746)

hectorcorrea:
When the current user's Mediaflux session times out while approving a project our code was leaving a Rails project record without a matching Mediaflux collection asset. 

I moved the code that deletes the orphan records to check at the very end of the process if we need to delete the orphan record. 

Closes #1800

### DIFF
--- a/app/operations/project_create.rb
+++ b/app/operations/project_create.rb
@@ -19,8 +19,8 @@ class ProjectCreate < Dry::Operation
       # Create the project in the Rails database
       project_metadata_json = RequestProjectMetadata.convert(request)
       # Link the request to the project
-      request.project_id = project.id
       project = Project.create!({ metadata_json: project_metadata_json })
+      request.project_id = project.id
       project.draft_doi
       project.save!
 


app/models/request.rb
```diff
@@ -144,4 +145,15 @@ def valid_uid?(uid, field)
         true
       end
     end
+
+    # If a request fails to be a approved we make sure there were not orphan
+    # project records left in our Rails database that do not have a matching
+    # project in Mediaflux (i.e. collection asset).
+    def cleanup_incomplete_project
+      project = Project.find_by_id(project_id)
+      if project && project.mediaflux_id.nil?
```
hectorcorrea, line 154: Notice that we only delete the project if it does not have a matching project in Mediaflux

app/operations/project_create.rb
```diff
@@ -22,6 +22,9 @@ def create_project_from_request(request)
       project.draft_doi
       project.save!
 
+      # Link the request to the project
```
carolyncole, line 25: @hectorcorrea should there be a request.save here so the request is saved before any error occurs?

spec/operations/project_create_spec.rb
```diff
@@ -34,12 +34,10 @@
 
     context "Failure cases" do
       it "returns a failure if the project cannot be saved to Mediaflux" do
-        expect do
-          result = described_class.new.call(request: invalid_request, approver: approver)
-          expect(result).not_to be_success
-          error_message = result.failure
-          expect(error_message).to include("Error saving project")
-        end.not_to change { Project.count }
+        result = described_class.new.call(request: invalid_request, approver: approver)
```
carolyncole, line 37: Could you change this to verify that the project still exists.  `end.to change { Project.count }.by(1)`

carolyncole (changes requested):
I'm a bit concerned about a couple changes.  I do not see a test of the request project id setting.  I doubt it will actually stick without a save of the request.

carolyncole (approved):
Thanks @hectorcorrea and @jrgriffiniii

Issue #1806, opened 2025-09-02
sidekiq should be auth protected


bess:
### Expected behavior
When I go to the `/sidekiq` path, I should need to be authenticated as a sysadmin or a developer

### Actual behavior
it's public and it's a security problem

### Acceptance criteria
- [ ] Sysadmins and superusers should be able to access the `/sidekiq` but no one else

PR #1808, opened 2025-09-02
Only developers and sysadmin can reach sidekiq dashboard

bess:
Ref #1806

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -31,8 +31,8 @@ class ApplicationController < ActionController::Base
   end
 
   # Render a 404 page for any undefined route
-  def render_404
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -30,6 +30,11 @@ def add_breadcrumb(name, path = nil)
     breadcrumbs << Breadcrumb.new(name, path)
   end
 
+  # Render a 404 page for any undefined route
+  def render_not_found
+    render file: "#{Rails.root.join('public', '404.html')}, layout: false, status: :not_found
```
carolyncole, line 35: I think the syntax error here is causing all the issues
```suggestion
    render file: "#{Rails.root.join('public', '404.html')}", layout: false, status: :not_found
```

Issue #1790, opened 2025-08-26
Allow any user to see their sponsored and managed projects


carolyncole:
### Expected behavior
If I am set as a sponsor or data manager on a project it should show up on my dashboard.

### Actual behavior
We are now allowing all users to be data sponsors and/ or data managers, but when loading projects for their dashboard we only show the projects they are data users for if they are not marked eligible for data manager or data sponsor https://github.com/pulibrary/tigerdata-app/blob/main/app/models/project.rb#L133-L140

### Steps to replicate

Create a project with a user set as the sponsor.  Check the Project.user_projects and see that the array returned is empty.

### Impact of this bug
User will have an invalid list of projects on their dashboard.

### Acceptance criteria
- [ ] If this change touches any of the files referenced in `.rubocop_todo.yml` create a new ticket to address the rubocop  todo.
- [ ] When a user is a sponsor of a project it should show on their dashboard
- [ ] When a user is a data manager of a project it should show on their dashboard

PR #1810, opened 2025-09-03
allow users to see sponsored and managed projects

leefaisonr:
ref #1790


carolyncole (approved):
I do not see why your PR has impacted the coverage.  I think this is OK...

PR #1814, opened 2025-09-04
Making the new_project_request_wizard_spec.rb run more consistently

carolyncole:
Also change how we lookup users from ruby code to a database query to make it faster (thought timing might have been part of the test issue) And removed pulls for the entire user list since it is no longer being utilized

I ran the test 10 times `for run in {1..10}; do rspec ./spec/system/new_project_request_wizard_spec.rb:136; done` and kept changing the sleep until the test was green for 2 runs of 10 in a row.  That is how I got `1.2` seconds.  I'm not sure if that will work for everyone, but hopefully...

It turns out debounce was not always sending the ajax request and getting a response before the half second mark.

### DIFF
--- a/spec/services/princeton_users_spec.rb
+++ b/spec/services/princeton_users_spec.rb
@@ -100,32 +100,38 @@ RSpec.describe PrincetonUsers, type: :model do
     end
   end
 
-  describe "#user_match?" do
+  describe "#user_list_query" do
     let(:user) { { uid: "sms98", name: "Sotomayor, Sonia" } }
-    let(:user_no_name) { { uid: "sms98" } }
+    let(:user_no_name) { { uid: "sms99", name: nil } }
+
+    before do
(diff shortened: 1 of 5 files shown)


spec/services/princeton_users_spec.rb
```diff
@@ -100,32 +100,38 @@
     end
   end
 
-  describe "#user_match?" do
+  describe "#user_list_query" do
     let(:user) { { uid: "sms98", name: "Sotomayor, Sonia" } }
-    let(:user_no_name) { { uid: "sms98" } }
+    let(:user_no_name) { { uid: "sms99", name: "" } }
```
hectorcorrea, line 105: @carolyncole we have users with a nil name in the database, that's why I had the name as nil in the tests. I am not sure how the new code will handle them.

bess (approved):

PR #1828, opened 2025-09-09
Send mail after a request is submitted, not saved

JaymeeH:
Changing mail to be sent after a request is submitted instead of whenit is saved


carolyncole (approved):
Thanks for the quick fix @JaymeeH

PR #1868, opened 2025-09-18
Fixes bug when there is a field without a Mediaflux label

hectorcorrea:
With the [introduction of the new fields](https://github.com/PrincetonUniversityLibrary/tigerdata-config/pull/223) (Project Purpose, HPC, SMB, Globus) we discovered that the frontend is not handling gracefully when a field is defined in Mediaflux without a label. 

In particular the Project Purpose field does not have a label and it was crashing the project creation process. This PR fixes that bug.


carolyncole (approved):
Thanks @hectorcorrea

PR #1888, opened 2025-09-22
Define metadata schema update process

cbentlerOIT:
Advances #1887 (Define the process for frontend / backend interoperability during TigerData schema updates.)


hectorcorrea (approved):
Looks good to me

Issue #1890, opened 2025-09-23
When the uid we get from LDAP is blank, do not make a user


bess:
### Expected behavior
When the uid we get from LDAP is blank, do not make a user. Instead, just report it to honeybadger.

### Acceptance criteria
- [x] A blank :uid field in the data we get from LDAP does not result in the creation of a User object with an empty :uid field

PR #1895, opened 2025-09-24
Improve error handling for malformed ldap entries

bess:
Ref #1890


carolyncole (approved):
I would have expected code when the user logs in to be modified based on the conversation yesterday, but this seems ok

Location I was expecting change...
https://github.com/pulibrary/tigerdata-app/blob/6c04b0e43f6e8efbb0205908ccad64823d71c12d/app/models/user.rb#L18

Issue #1860, opened 2025-09-17
System spec for Space Ghost increment


kelynch:
A system spec is created to test the functionality in the Space Ghost increment.

Acceptance Criteria
- [x] Create a Space Ghost Spec
- [x] A user should be able to login to get to the project request wizard. (not a sysadmin, datamanager, or datasponsor)
- [x] All of the fields on the drupal form exist on the project request wizard (https://tigerdata.princeton.edu/get-started)

Implementation notes

- [x] The data users table will need to be integrated into the wizard for us to test that feature in the system spec. Ref #1511

precillap:
Is the "additional information" section that's on the current druple shown below required to be in the wizard?

<img width="1108" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNXFNQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6454505221b803022ec2979d291ae3ff0de499fb/Screenshot%202025-09-29%20at%202.38.59%E2%80%AFPM.png" alt="Screenshot 2025-09-29 at 2.38.59 PM.png" />

PR #1916, opened 2025-09-29
Tests that all drupal form fields exist on the new project wizard

precillap:

ref #1860


spec/system/space_ghost_spec.rb
```diff
@@ -56,7 +58,13 @@
       select another_user_str + "\u00A0", from: "user_find"
       click_on "Add Users"
 
-      # next step is to click on "add user" then figure out how to load in the user for the druple
+      # This part of the code is going through the rest of the fields of the wizard form to review and submit the project request
+      click_on "Next"
+      expect(page).to have_content("Enter the storage and access needs for your project.")
+      click_on "Next"
+      expect(page).to have_content("Review")
```
carolyncole, line 65: FYI I think review is on every page (it's in the side bar).  A better check for the review page would be to test for below
```suggestion
      expect(page).to have_content("Take a moment to review your details ")
```

carolyncole (changes requested):
One last change and I think it looks good.  You found so many bugs in the code writing this test.  Thank you!

PR #1943, opened 2025-10-01
Only run the vite build process once

carolyncole:


carolyncole:
This seems to be slower than the original build time 11 minutes vs 8

Issue #1615, opened 2025-07-17
Cleaning previous test data from Mediaflux


hectorcorrea:
With PR https://github.com/pulibrary/tigerdata-app/pull/1572 we are now creating test data in Mediaflux under the same root for Development as we do for Testing which is under `/princeton/tigerdata`. This is because the `tigerdata.project.create` service is hard-coded this way.

Because of this we are not cleaning Mediaflux before we run the tests. Towards the end of PR #1572 I managed to get all the tests to create their data under `/princeton/tigerdata/rspec` which would make it it possible for us to delete all test data from Mediaflux without blowing up data added for development.

Revisit if we can indeed delete all test data before running the rspec suite. See https://github.com/pulibrary/tigerdata-app/pull/1572/files#diff-63f3e9f46c3167f862873bd7e9b9385231b80d55485e9127bf706296d4a7ddf6L18

PR #1951, opened 2025-10-02
cleanup test data from mediaflux

JaymeeH:
ref #1615


spec/models/mediaflux/iterator_destroy_request_spec.rb
```diff
@@ -2,7 +2,7 @@
 require "rails_helper"
 
 RSpec.describe Mediaflux::IteratorDestroyRequest, connect_to_mediaflux: true, type: :model do
-  let(:mediaflux_url) { "http://0.0.0.0:8888/__mflux_svc__" }
+  let(:mediaflux_url) { Mediaflux::Request.uri.to_s }
```
JaymeeH, line 5: This hard coded url was causing the test to fail when running against `mflux_ci`

bess (approved):

Issue #2054, opened 2025-10-17
Error showing project file list in Show page


hectorcorrea:
We got this error in Honeybadger for a user on the Show page:

https://app.honeybadger.io/projects/113559/faults/124366788

Chuck thinks this might be because the files for this particular project are on a location with long retrieval times (something to do with ACL calculation)

hectorcorrea:
Bumped the timeout to 4 minutes manually in Production (see changes in #2056) and it did not address the issue with the "PNI Archive" project. Chuck ran a query directly from aterm on that project and the first `asset.query.iterate` takes 18 minutes 😱 

Chuck also noticed that the second largest project project (cee/hydrology) runs perfectly fine, the `asset.query.iterate` completes in 0.039 seconds. 

So there is something else going on with the PNI Archive. More to come.

PR #2056, opened 2025-10-17
Temporarily bumping the read timeout

hectorcorrea:
Temporarily bump the read timeout value to 4 minutes and see if that works.

Part of #2054


hectorcorrea:
Manually applied this to production and did not address the issue (see notes in #2054). Closing until we know if we need to tweak the timeout values.

Issue #2038, opened 2025-10-15
Honeybadger alert for projects returned where a user has no defined role


kelynch:
Add a Honeybadger alert for when we get projects back from Mediaflux for dashboard display where the user has access to a project but has none of the three roles (data sponsor, data manager, data user) that reflect access.

## Acceptance criteria

- [ ] Check the data manager, data sponsor, and data users fields from Mediaflux for the user's net ID
- [ ] If not found in any of those fields, log the net ID and Mediaflux collection ID in Honeybadger

PR #2057, opened 2025-10-17
Dashboard displays projects straights from Mediaflux

### LINKED ISSUE #2037: Dashboard displays projects the user has access to in Mediaflux
Using `asset.query` to return all projects that a user has access to in Mediaflux so that the dashboard can be populated.

Example query:
`asset.query :where xpath(tigerdata:project/ProjectID) has value`

Return the following values from Mediaflux:
* Collection ID from Mediaflux
* Project title
* User role (data sponsor, data manager, data user)
* Quota information (total quota and consumed)

The following values come from the web application and should stay that way for the moment:
* Project status (do we need this?)
* "Standard" is hard-coded in the web application
* "Downloads" comes from the activity in the web application
* "Active/not yet active" comes from the activity in the web application

## Acceptance criteria

- [ ] The expected values are returned from Mediaflux with a call to `asset.query`
- [ ] The Mediaflux values replace the Rails values in the dashboard
- [ ] Rails projects no longer display on the dashboard
- [ ] All tests pass

hectorcorrea:
The list of projects displayed in the **Dashboard** now comes straight from Mediaflux. 

The code in this PR is a bit messy because the `ProjectDashboardPresenter` inherits from the `ProjectShowPresenter` and this PR only updates the project information in the Dashboard (not on the project Show page). Therefore there a few extra `if/else` statements to account for the two kind of data that the presenter might receive. I am hoping these `if/else` statements will be removed once we update the Project Show page (https://github.com/pulibrary/tigerdata-app/issues/2039) to also fetch its data directly from Mediaflux.

Closes #2037 (and #2038)

Example in QA of the Dashboard showing **only my Projects** in the Projects tab and **all the projects** in the Admin tab (I am an Administrator in QA):

mine

not-mine

### DIFF
--- a/app/views/requests/index.html.erb
+++ b/app/views/requests/index.html.erb
@@ -1,5 +1,5 @@
 <h1>New Project Requests</h1>
-<h2>Draft Requests</h2>
-<%= render partial: "request_list", locals: {request_list: @draft_requests } %>
 <h2>Submitted Requests</h2>
 <%= render partial: "request_list", locals: {request_list: @submitted_requests } %>
+<h2>Draft Requests</h2>
+<%= render partial: "request_list", locals: {request_list: @draft_requests } %>


app/models/project.rb
```diff
@@ -96,27 +96,20 @@ def in_mediaflux?
...
-
-  def self.sponsored_projects(sponsor)
-    Project.where("metadata_json->>'data_sponsor' = ?", sponsor)
-  end
-
-  def self.managed_projects(manager)
-    Project.where("metadata_json->>'data_manager' = ?", manager)
+    all_projects(user).select do |project|
+      project[:data_manager] == user.uid || project[:data_sponsor] == user.uid || project[:data_users].include?(user.uid)
+    end
```
hectorcorrea: Filter the projects that came from Mediaflux. 

Notice that Mediaflux filters projects that the user has access to on the backend, in here we just make sure the user has a roles on the project (manager, sponsor, data user.) This extra filtering is necessarily because system administrators have access to other projects in Mediaflux even when they don't have a role in them.

app/presenters/dashboard_presenter.rb
```diff
@@ -6,14 +6,24 @@ def initialize(current_user:)
...
+    # Short-cut to prevent an extra call to Mediaflux for non-admins
+    # which are the large majority of our users.
     return [] unless current_user&.eligible_sysadmin?
 
-    @all_projects ||= Project.all_projects.map { |project| ProjectDashboardPresenter.new(project) }.sort_by(&:updated_at).reverse
+    @all_projects ||= Project.all_projects(current_user)
+                             .map { |project| ProjectDashboardPresenter.new(project) }
+                             .select(&:project_in_rails?)
+                             .sort_by(&:updated_at)
+                             .reverse
```
JaymeeH: NIce shortcut!

Issue #2065, opened 2025-10-21
Flaky test: NewProjectWizard::ReviewAndSubmitController#show a signed in user a developer the production environment shows the form


bess:
Why is this test flaky? 

See this test run: https://app.circleci.com/pipelines/github/pulibrary/tigerdata-app/6065/workflows/dc53d544-c66a-4e2c-8c55-77241726c59c/jobs/13709

<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeFBMQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ee26223054adb0533a983dca29adbe0ecbfee125/Screenshot%202025-10-21%20at%2012.16.00%E2%80%AFPM.png" alt="Screenshot 2025-10-21 at 12.16.00 PM.png" />

PR #2070, opened 2025-10-21
Remove tests for special behavior in production

bess:
...because everything should match now anyway, and these tests are flaky and order dependent.

The order dependency can be re-created with this command:
> be rspec --seed 28249 spec/controllers/new_request_wizard/

Also, there are sometimes errors related to email addresses not being unique, so this ensures email addresses are always unique. 

Ref #2065


carolyncole (changes requested):
Actually it will not all match.  The developers are not sysadmins in production and never will be.

PR #2098, opened 2025-10-27
Upgrade to ruby 3.4.7

carolyncole:


carolyncole:
This is too old to keep open

PR #2128, opened 2025-10-30
Project Show page now displays metadata straight from Mediaflux

### LINKED ISSUE #2039: Project details page displays information from Mediaflux
Using `asset.get` and passing the Mediaflux collection ID, return all information on the Project Details page from Mediaflux, when accessing the Project Details page

Return the following data from Mediaflux:
* Project title
* Mediaflux collection ID
* Project ID (the DOI)
* Description
* Quota (total and consumed)
* Data Sponsor
* Data Manager
* Data User(s)
* Department(s)
* Project directory (parent and project directory)
* Project purpose
* Estimated number of files
* High perfomrance computing
* Network file shares
* Globus transfers

The following information is provenance info and comes from the Rails project representation and should stay as-is:
* Storage performance expectations (requested and approved) hardcoded to "Standard" -- should this be removed?  We don't appear to be using it.
* Requested storage capacity
* Approved storage capacity
* Mediaflux errors that happened when the project was created (only visible to sysadmins and developers) -- no changes should be needed for this

## Acceptance criteria

- [ ] Get back all information required from Mediaflux
- [ ] Values from Mediaflux are displayed in the project details page instead of from the Rails project
- [ ] Provenance information is displated from the Rails project
- [ ] Tests pass

hectorcorrea:
Updated project show page (Show and Details) to fetch the project metadata straight from Mediaflux.

Notice that I am performing most of the changes at the Presenter level (rather than at the Model) because we need a valid user session to fetch from Mediaflux and the presenter seems like a better place for this.

NOTE: There are other places in the code that still reference the original project metadata stored in our Rails record (e.g. mailers) and we should refactor that code too to fetch the data from Mediaflux but I am trying to keep this PR as small as possible. See issue: https://github.com/pulibrary/tigerdata-app/issues/2132 for details.

Closes #2039

### DIFF
--- a/spec/controllers/projects_controller_spec.rb
+++ b/spec/controllers/projects_controller_spec.rb
@@ -3,8 +3,8 @@ require "rails_helper"
 
 RSpec.describe ProjectsController, type: ["controller", "feature"] do
   let!(:sponsor_and_data_manager) { FactoryBot.create(:sponsor_and_data_manager, uid: "tigerdatatester", mediaflux_session: SystemUser.mediaflux_session) }
-  let(:request) { FactoryBot.create :request_project, data_manager: sponsor_and_data_manager.uid, data_sponsor: sponsor_and_data_manager.uid }
-  let(:project) { request.approve(sponsor_and_data_manager) }
+  let(:project_request) { FactoryBot.create :request_project, data_manager: sponsor_and_data_manager.uid, data_sponsor: sponsor_and_data_manager.uid }
+  let(:project) { project_request.approve(sponsor_and_data_manager) }
 
   before do
     Affiliation.load_from_file(Rails.root.join("spec", "fixtures", "departments.csv"))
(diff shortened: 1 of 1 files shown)


app/models/project.rb
```diff
@@ -112,20 +112,10 @@ def self.all_projects(user)
     request.results
   end
 
-  def user_has_access?(user:)
-    return true if user.eligible_sysadmin?
-    metadata_model.data_sponsor == user.uid || metadata_model.data_manager == user.uid ||
-    metadata_model.data_user_read_only.include?(user.uid) || metadata_model.data_user_read_write.include?(user.uid)
-  end
-
```
hectorcorrea: Moved to the presenter to use the Mediaflux data (rather than our fields on the Rails database)

spec/controllers/projects_controller_spec.rb
```diff
@@ -196,12 +206,12 @@
...
         let(:user) { User.find_by(uid: project.metadata_model.data_sponsor) }
 
-        it "does not contact mediaflux" do
+        it "does contact mediaflux" do
           allow(Mediaflux::QueryRequest).to receive(:new).and_call_original
 
           get :show, params: { id: project.id }
 
-          expect(Mediaflux::QueryRequest).not_to have_received(:new)
+          expect(Mediaflux::QueryRequest).to have_received(:new)
```
hectorcorrea: This is the new behavior since we are fetching the metadata from Mediaflux

bess (changes requested):
Overall this looks like really great work! Thank you! I don't think it's a good idea to xit these tests, though. Could we take a little more time and figure out why they're failing?

hectorcorrea:
@bess I am looking into the tests that I `xit` on this PR as part of a separate ticket: https://github.com/pulibrary/tigerdata-app/issues/2142 

If that's the only issue could we merge this PR? It's too big and the longer it stays open the longer I risk merge conflicts.

Issue #2100, opened 2025-10-27
Spike: Explore implementing a Java Plugin/Docker Build


jrgriffiniii:
This aims to build upon the advancements being made by colleagues in Research Computing (@mduchnowski-pu) in order to extend the features and functionality of the Mediaflux code base.

## Acceptance criteria
- [x] Augments the process at https://github.com/PrincetonUniversityLibrary/tigerdata-config
- [x] Ensure this documentation works: https://github.com/PrincetonUniversityLibrary/tigerdata-config/blob/main/docs/iterate_locally.md
- [x] Get a .jar file from Matt D and add it to our tigerdata-config build process such that it's installed in the right place for the final docker image
- [x] We will know this is working when we can log into a mediaflux_dev docker container and call the `tigerdata.trivial` service to reverse a string. Write an automated test in tigerdata-app that does this.

jrgriffiniii:
https://github.com/PrincetonUniversity/td-mf-java-plugin/pull/1 is WIP but is advancing some of the fundamental work required for integration with GitHub Actions

jrgriffiniii:
the `arc.xodb` package is required to build and test plugins outside of the Mediaflux environment, and appears to only be available in the `aserver.jar`.

PR #2160, opened 2025-11-06
Java plugin proof of concept

bess:
Ref #2100

### DIFF
--- a/app/models/mediaflux/string_reverse.rb
+++ b/app/models/mediaflux/string_reverse.rb
@@ -14,7 +14,17 @@ module Mediaflux
     # Specifies the Mediaflux service to use when querying Mediaflux.
     # @return [String]
     def self.service
-      "tigerdata.trivial :string #{@string}"
+      "tigerdata.trivial"
     end
+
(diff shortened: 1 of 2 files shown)


app/models/mediaflux/string_reverse.rb
```diff
@@ -0,0 +1,20 @@
...
+    # @param session_token [String] the API token for the authenticated session
+    def initialize(string:, session_token:)
+      super(session_token: session_token)
+      @string = string
+    end
+
+    # Specifies the Mediaflux service to use when querying Mediaflux.
+    # @return [String]
+    def self.service
+      "tigerdata.trivial :string #{@string}"
```
hectorcorrea: The service should be just the name of the service: `tigerdata.trivial`, we pass the parameter as a separate block since Mediaflux expects them in a different part of the XML when it receives the request. 

Take a look at this for an example:  https://github.com/pulibrary/tigerdata-app/blob/main/app/models/mediaflux/asset_metadata_request.rb#L47-L53

PR #2164, opened 2025-11-06
Fixes bug in check for user_has_access for data users

hectorcorrea:
There was a bug in the presenter code that checks whether a user has access to a project, the code was treating the `data_users` as a list of `uid`s but that list is now a list of `User`s. I probably introduced this bug when I refactored the code that fetches data straight from Mediaflux. 

This PR fixes the issue and adds a test for it.

### DIFF
--- a/app/models/mediaflux/string_reverse.rb
+++ b/app/models/mediaflux/string_reverse.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+module Mediaflux
+  class StringReverse < Request
+    # This is only here to prove that we can use a mediaflux service that is provided by
+    # a java plugin.
+
+    # Constructor
+    # @param session_token [String] the API token for the authenticated session
+    def initialize(string:, session_token:)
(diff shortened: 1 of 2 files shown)


app/presenters/project_show_presenter.rb
```diff
@@ -148,7 +148,8 @@ def quota_percentage(session_id:)
 
   def user_has_access?(user:)
     return true if user.eligible_sysadmin?
-    data_sponsor.uid == user.uid || data_manager.uid == user.uid || data_users.include?(user.uid)
+    return true if data_sponsor.uid == user.uid || data_manager.uid == user.uid
+    data_users.find { |data_user| data_user.uid == user.uid } != nil
```
carolyncole: I think the include? easier to read.  Could you change the line to:
```suggestion
     data_users.map(&:uid).include?(user.uid)
```

Or I wonder if data_users.include?(user) would work?

PR #2177, opened 2025-11-10
Handles access error gracefully

### LINKED ISSUE #2155: Handle gracefully when user does not have access to a project
If a user has link to a project that they don't have access to (e.g. https://tigerdata-app.princeton.edu/projects/1234/details) we run into an error because the user does not have access to the project and the throw an exception which results in the unfriendly "Something went wrong" message, we should handle this more gracefully.

https://app.honeybadger.io/projects/113559/faults/125035127

hectorcorrea:
This error only happened if a user got a deep link to a project that they didn't have access to so it is not high visibility, but now we handle the issue gracefully rather than showing the "something went wrong" page.

Closes #2155

### DIFF
--- a/app/javascript/entrypoints/titleCopySaveExit.js
+++ b/app/javascript/entrypoints/titleCopySaveExit.js
@@ -2,7 +2,7 @@
 export function titleCopySaveExit() {
   // Disable or enable the confirm button depending on if the title is empty
   function verify() {
-    if (document.getElementById('project_title_exit').value === '') {
+    if (document.getElementById('project_title_exit').value.trim() === '') {
       document.getElementById('save-exit-confirm').disabled = true;
     } else {
       document.getElementById('save-exit-confirm').disabled = false;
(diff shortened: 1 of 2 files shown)


app/presenters/project_show_presenter.rb
```diff
@@ -148,7 +148,7 @@ def quota_percentage(session_id:)
 
   def user_has_access?(user:)
     return true if user.eligible_sysadmin?
-    data_sponsor.uid == user.uid || data_manager.uid == user.uid || data_users.map(&:uid).include?(user.uid)
+    data_sponsor&.uid == user.uid || data_manager&.uid == user.uid || data_users.map(&:uid).include?(user.uid)
```
hectorcorrea: This is really the fix: if the user does not have access to a project we cannot fetch the `data_sponsor` and we were getting into a `nil` exception. Now the code will return false which will prevent the access to the project.

JaymeeH (approved):
Thanks @hectorcorrea!

Issue #2147, opened 2025-11-05
Deduplicate departments on new project request submission


kelynch:
References https://github.com/pulibrary/tigerdata-app/issues/2134 

When a request is submitted, deduplicate the department field and only save each unique department value once in the array.

## Acceptance criteria

- [ ] A request with duplicate departments only saves each department once when it is submitted
- [ ] There are tests for this

PR #2229, opened 2025-11-20
Deduplicate department field from requests

JaymeeH:
ref #2147

### DIFF
--- a/app/controllers/request_wizards_controller.rb
+++ b/app/controllers/request_wizards_controller.rb
@@ -106,15 +106,13 @@ class RequestWizardsController < ApplicationController
     end
 
     # Only allow a list of trusted parameters through.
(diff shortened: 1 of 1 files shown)


app/controllers/request_wizards_controller.rb
```diff
@@ -106,13 +106,15 @@ def save_request
     end
 
     # Only allow a list of trusted parameters through.
+    # rubocop:disable Metrics/MethodLength
```
carolyncole:
```suggestion
```

carolyncole (commented):
Thanks @JaymeeH.  I would have preferred you not to ignore rubocop (See my suggestions) as a default and instead extract a concept method.  I'm not going to block this PR, but I'm also going to let someone else bless it.

PR #2281, opened 2025-12-05
Allow underscores as valid characters in the project directory

### LINKED ISSUE #2280: Allow undescores in the path
We currently don't allow underscores in the project directory and we should.

This is the code that converts them from underscores to dashes: https://github.com/pulibrary/tigerdata-app/blob/main/app/models/project_metadata.rb#L52

hectorcorrea:
Per a discussion with Andrew (RC) and @matthewjchandler we should accept underscores in the path.

Closes #2280

### DIFF
--- a/app/assets/stylesheets/_dashboard.scss
+++ b/app/assets/stylesheets/_dashboard.scss
@@ -331,6 +331,9 @@
       justify-content: center;
       align-items: center;
       gap: 0.75rem;
+      @media (min-width: 1280px) {
+        width: 20rem;
+      }
 
       .storage {
         display: flex;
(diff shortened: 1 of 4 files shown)


app/models/project.rb
```diff
@@ -253,11 +253,6 @@ def project_directory_pathname
       end
     end
 
-    # Ensure that the project directory is a valid path
-    def safe_directory(directory)
-      Project.safe_directory(directory)
-    end
-
```
hectorcorrea: This method was unused (and thank goodness for that because it would have been an infinite recursive call!)

leefaisonr (approved):
Thanks @hectorcorrea

Issue #2310, opened 2025-12-15
Review and consolidate project request tests


kelynch:
Review and consolidate tests around users creating project requests.  We currently have Skeletor-specific tests, Space Ghost-specific tests, and more general “user can create a new request” tests.  Review and consolidate these into one set of uniform tests that fully test project requests for each of the three user types (sysadmin user, researcher user, tester-trainer user).

PR #2339, opened 2026-01-05
Consolidate project request tests

JaymeeH:
ref #2310


spec/system/new_project_request_wizard_spec.rb
```diff
@@ -1,443 +1,579 @@
...
+          expect(page).not_to have_content "Draft request saved automatically"
+          request = Request.last
+          expect(request.project_title).to eq("Dashboard Redirect Test")
+        end.to change { Request.count }.by(1)
+      end
     end
   end
 end
+
+describe "#file_list", integration: true do
```
carolyncole: Curious why file list got added in here.  Seem like it is unrelated to the new request wizard

Issue #2326, opened 2025-12-18
Remove the temporary welcome banner


matthewjchandler:
We started with the banner upon launch in early 2026, and we don't need it anymore.

<img width="1408" height="1238" alt="Image" src="https://github.com/user-attachments/assets/ad4e6139-a389-4836-a8fb-da41a3dc608f" />

When we have a new message/need for a banner, we will create a new ticket for that.

### Acceptance Criteria

- [ ] The banner does not display for users in any environment:
  - [ ] QA
  - [ ] Staging
  - [ ] Production

PR #2345, opened 2026-01-08
remove the temporary welcome banner

JaymeeH:
ref #2326

### DIFF
--- a/config/banner.yml
+++ b/config/banner.yml
@@ -3,8 +3,8 @@ development:
   title:
   body:
 test:
-  title:
-  body:
+  title: Welcome to the <i>test</i> site
+  body: This is a <b>test</b> <a href="mailto:fake@princeton.edu">message.</a>
 staging:
   title:
   body:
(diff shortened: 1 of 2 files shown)


bess (changes requested):
@JaymeeH Are we being asked to remove the entire banner feature? Or just remove the banner as currently displayed in those environments? 

If we're just removing the banner from the environments (which is what I'd prefer) I think we can just remove the message from this config file: https://github.com/pulibrary/tigerdata-app/blob/main/config/banner.yml

That lets us keep the banner functionality in place so we can quickly restore it when needed. It's been very helpful on our other systems and I don't think we have a replacement feature for this yet. Glad to be corrected if I'm wrong!

bess:
@JaymeeH I'm so sorry, I'm not trying to be picky, but this still removes the feature. Could you please leave the config in place for the test environment and then you can leave the test for the feature in place too? You could change the header on the test to "it has a banner when there is a banner configured" or something?

PR #2365, opened 2026-01-21
Quota breakdown (first pass)

hectorcorrea:
First pass at calling the new quota plug-in and displaying the quota breakdown in the Project Details page if and only if the user passes a specific query string parameter.


carolyncole (approved):
Thanks @hectorcorrea! I wonder if a next step would be to show this only for sysadmins devs and tester trainers.

PR #2406, opened 2026-02-12
rename RequestController to NewProjectRequestController

carolyncole:
Also move path from /requests to /new_project_requests


leefaisonr (approved):
Thanks @carolyncole

PR #2411, opened 2026-02-13
Fixed copy button on project details page

precillap:


app/javascript/entrypoints/copyPastePath.js
```diff
@@ -74,7 +74,7 @@ export function copyPastePath() {
   });
 
   /* Basic Details section */
-  $('#copy-project-path-button-basic').on('click', (event) => {
+  $('.copy-project-path-button-basic').on('click', (event) => {
```
hectorcorrea: Why are we changing the CSS Selector here from `#` (id) to `.` (class) ?

Issue #1361, opened 2025-04-07
[tigerdata-app] Checkmk should monitor expiration of development license


bess:
The mediaflux development license expires annually, and we should get an alert a month ahead of time. Can checkmk look at the mediaflux license file, or a hardcoded value, so we can get an alert?

carolyncole:
We have information on when the license expires in github and we validate the license on every commit.  This means we do not just get random failures.  This is good enough.

PR #2414, opened 2026-02-18
Output the current license information to the command line on commit

carolyncole:
also put the expiration date in circle-ci output

It will raise an exception if the license is expired.  CI will not pass with an expired license (this is just an example)

Screenshot 2026-02-18 at 12 47 11 PM

refs #1361


carolyncole:
Thanks for catching the name typo @hectorcorrea.  I updated the PR.

hectorcorrea (approved):

PR #2452, opened 2026-03-05
Use tigerdata.describe to fetch Mediaflux information

hectorcorrea:
Changed the code to use the information from the `tigerdata.describe` plugin to fetch Mediaflux server information (see screenshot below).

Notice that I also changed the "mediaflux_info" URL to "session-info" since (1) "mediaflux_info" was too close to the "mediaflux_passthru" and "mediaflux_extra" that we use for CAS authentication and I wanted to prevent users from accidentally going to the wrong place and (2) we display more than just Mediaflux information on that page and "session-info" seems more appropriate.

Screenshot 2026-03-05 at 2 12 22 PM

## NOTE 
~~I am keeping this PR in draft mode until we have the Java plugin deployed to production. If we deploy earlier the frontend will stop displaying the server information since the new code expects the `tigerdata.describe` plugin to be available, not catastrophic but not great either.~~ 

The Java plugin has now been rolled out to production so we are ready to merge this.

### DIFF
--- a/app/assets/stylesheets/_projects.scss
+++ b/app/assets/stylesheets/_projects.scss
@@ -302,9 +302,13 @@
       color: $gray-60;
     }
   }
+  .detailed-quota-subtypes-container {
+    display: flex;
+  }
   .detailed-quota-subtypes-free-space {
-    display: ruby;
-    margin-left: 8rem;
+    display: flex;
+    margin-left: auto;
+    gap: 0.5rem;
     .subtype-size-free-space {
       color: $gray-60;
       font-size: 0.875rem;


app/models/mediaflux/tigerdata_describe_request.rb
```diff
@@ -0,0 +1,28 @@
...
+    # Constructor
+    # @param session_token [String] the API token for the authenticated session
+    def initialize(session_token:)
+      super(session_token: session_token)
+    end
+
+    # Specifies the Mediaflux service to use when querying Mediaflux.
+    # @return [String]
+    def self.service
+      "tigerdata.describe"
```
hectorcorrea: This is the new plugin that @mduchnowski-pu wrote that give us the information.

app/views/session_info/index.html.erb
```diff
@@ -0,0 +1,48 @@
...
+  <dd><%= @mediaflux_info[:name] || "UNKNOWN" %></dd>
+
+  <dt>Server Version:</dt>
+  <dd><%= @mediaflux_info[:server_version] %></dd>
+
+  <dt>TigerData Config:</dt>
+  <dd><%= @mediaflux_info[:tigerdata_config_version] %></dd>
+
+  <dt>TigerData Plugin:</dt>
+  <dd><%= @mediaflux_info[:tigerdata_plugin_version] %></dd>
```
hectorcorrea: And with that new plugin we now have access this kind of information 😍

Issue #2422, opened 2026-02-24
Enable form submit action for Request More Storage form


kelynch:
Part of https://github.com/pulibrary/tigerdata-app/issues/2378

Enable the submission action on the form implemented in https://github.com/pulibrary/tigerdata-app/issues/2421

## Behavior

Completing the required fields and clicking "Submit" will trigger two simultaneous actions:

1. For the user, a confirmation will be displayed as a different modal, appearing instead of the request form:

<img width="534" height="233" alt="Image" src="https://github.com/user-attachments/assets/d31f6288-f139-4b97-969f-8b0f5cc3e74e" />

Clicking "Back to Storage Overview" will return the user to the overview modal (we won't implement a badge for pending status right now).

2. For the request to be processed, submitting the completed form will also send an email with the following properties:

- To: tigerdata@princeton.edu
- From: {NetID email of the user who submitted the form}
- Subject: "Storage Increase Request Ready for Review"
- Body:

> A storage increase request has been created and is ready for review. Here are the request details:
> 
> Project Title: {insert title}
> Project Directory: {insert directory path}
> Project ID: {insert DOI}
> Data Sponsor: {insert sponsor name and NetID}
> Data Manager: {insert manager name and NetID}
> Current Storage Capacity: {insert total quota from Mediaflux}
> Current Usage: {insert total usage from Mediaflux}
> - Project Files: {insert subtotal usage for project files}
> - Old Versions: {insert subtotal usage for old versions}
> - Recycle Bin: {insert subtotal usage for recycle bin}
> 
> Requested Storage Capacity: {insert amount from form}
> Justification: {insert value from form}
> Growth Expectations: {insert value from form}
> Date Needed: {insert value from form as YYYY-MM-DD}

Clicking the "X" button in any modal will return the user to the Project page.

See the Figma [Prototype](https://www.figma.com/proto/QthyYyx8IHYzUTGcHf0rdy/TigerData-UI-TM-DP?page-id=932%3A48590&node-id=2590-4163&viewport=-3138%2C-4705%2C0.16&t=dkxWNMGqUReHD6QC-1&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=1372%3A16919&show-proto-sidebar=1)

### Link to Figma (Dev Mode)

https://www.figma.com/design/QthyYyx8IHYzUTGcHf0rdy/TigerData-UI-TM-DP?node-id=2590-3824&m=dev

## Implementation notes

[...]

PR #2509, opened 2026-03-23
Enable form submit action for storage request form

JaymeeH:
ref #2422


carolyncole (changes requested):
@JaymeeH I'm noting there are no tests for the mail actually going out, and/or tests for the contents of the email.  I see in your video that the modals work properly, but if the mail does not get sent then there is no actual record that the user made the request.

PR #2554, opened 2026-04-01
Updating tigerdata config to v0.37.0

carolyncole:
also adds in the testing users to the users loaded on local machines


leefaisonr (approved):
Thanks @carolyncole

Issue #2513, opened 2026-03-25
Change the current File Details Display to Display the top level only


carolyncole:
Based on User Story 31 we will display 100 items at each level to allow the user to see all of their items (99% of folders have less than 100 items in each folder currently in the system)

<img width="3190" height="11888" alt="Image" src="https://github.com/user-attachments/assets/e1154334-dad2-4fc9-bdb8-ba3174f37261" />

Our current display shows all levels paginated.  We need to update the default display to only show the top level and show all items without pagination.

## Figma link
https://www.figma.com/design/QthyYyx8IHYzUTGcHf0rdy/TigerData-UI-TM-DP?node-id=5430-2923&t=MXd3dQS7iRkyCsay-4

## Implementation details

Latest Java plugin implements [sample projects](https://github.com/PrincetonUniversityLibrary/tigerdata-config/issues/334) that are propagated across all environments under`rdss-query`, available if you pull down the latest Docker image.

## Hector's notes

See multiple comments below.

## Acceptance Criteria:

- [ ] When the File Details tab is opened the top level of the project (top folder) is displayed by default
- [ ] The top level will be truncated at 100 items
- [ ] No iterator is needed now, but we will in the future
- [ ] The query to mediaflux only retrieves the top level (see the prototype for for the asset.query)
- [ ] We will show the items in the order mediaflux returns them
- [ ] There are tests for showing up to 100 without sub folders, using the new projects in tigerdata-config (see Implementation details above)

User story document:
https://docs.google.com/document/d/1DMNF_8lWlt6GjDKCBTZd5-ZTRe7RsJQ-cRnBIUDXEvk/edit?tab=t.0

kelynch:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/gh/pulibrary/tigerdata-app/2513?planning-poker) with Zenhub @carolyncole @JaymeeH @jrgriffiniii @leefaisonr @precillap

PR #2561, opened 2026-04-03
Change the current File Details screen for the browse files Display to Display the top level only

leefaisonr:
ref #2513


carolyncole (approved):
Thanks @leefaisonr @kelynch

PR #2565, opened 2026-04-07
Updating rack for dependencies

kelynch:


bess (changes requested):
The rack part looks fine but I'm worried about the change to mediaflux-dev-license-expiration.txt. I don't know enough about what that file is and what Jaymee did earlier to fix the license to know whether it's going to cause a problem. Seems safer to remove it?

Issue #2540, opened 2026-03-26
PULCache class is untested


carolyncole:
### What maintenance needs to be done?

This is part of our health page.  The system will seem down if this code breaks.  This code should be tested

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
When we upgrade dependencies we should be easily able to tell if this code breaks

### Acceptance criteria

 - [ ] write an automated test for [PULCache](https://coveralls.io/builds/78473766/source?filename=app%2Fservices%2Fpul_cache.rb#L7)

PR #2580, opened 2026-04-10
Cover PULCache in test suite

JaymeeH:
ref #2540

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -30,7 +30,7 @@ Rails.application.configure do
   # Show full error reports and disable caching.
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
-  config.cache_store = :null_store
+  config.cache_store = :memory_store
 
   # Raise exceptions instead of rendering exception templates.
   config.action_dispatch.show_exceptions = false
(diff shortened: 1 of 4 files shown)


config/environments/test.rb
```diff
@@ -27,10 +27,11 @@
     "Cache-Control" => "public, max-age=#{1.hour.to_i}"
   }
 
-  # Show full error reports and disable caching.
+  # Show full error reports
+  # Allow caching to test PULCache.
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
-  config.cache_store = :null_store
+  config.cache_store = ENV["CI"].present? ? :null_store : :memory_store
```
carolyncole, line 34: What does this do differently than. the mock in the test?  Are both needed?

carolyncole (approved):
Thanks @JaymeeH 

Not going to block this going in... Do we need both the config change and the test mock? (a curiosity thing ...)

config/environments/test.rb
JaymeeH: This allowed the rest of the test suite that requires `null_store` to run without error in circleci, in addition to allowing us to run the test locally. Since we're mocking in the specific test, we don't necessarily need the config change anymore.

PR #2582, opened 2026-04-13
Removing Error classes that are unused

carolyncole:
Remove MetadataError && MissingMetadata
Update spec to no longer refrence class in title


JaymeeH (approved):
Thanks @carolyncole

PR #2589, opened 2026-04-14
Use null store by default for test config

### LINKED ISSUE #2540: PULCache class is untested

JaymeeH:
ref #2580 
closes #2540

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -31,7 +31,7 @@ Rails.application.configure do
   # Allow caching to test PULCache.
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
-  config.cache_store = ENV["CI"].present? ? :null_store : :memory_store
+  config.cache_store = :null_store
 
   # Raise exceptions instead of rendering exception templates.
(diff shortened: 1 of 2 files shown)


config/environments/test.rb
```diff
@@ -31,7 +31,7 @@
   # Allow caching to test PULCache.
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
-  config.cache_store = ENV["CI"].present? ? :null_store : :memory_store
```
JaymeeH, line 34: reverts the config change made in #2580

carolyncole (approved):
Thanks @JaymeeH

PR #2599, opened 2026-04-16
Do not reload test projects on non test systems

carolyncole:
We do not have access to the import, so if we delete the test projects they will not get reloaded


bess (approved):
Good catch, thank you!

PR #2624, opened 2026-04-23
Upgrading to tigerdata-config 0.40.0

carolyncole:
Updates to a newer version of the java plugin
Requires us to load the test projects on server start


carolyncole:
Close in favor of version 0.41.0

PR #2633, opened 2026-04-27
Refine integration suite

JaymeeH:
we only need the integration tests (`mflux_ci` job) to run when tigerdata-config is deployed to `mflux-ci1`. This will speed up our test suite since we are only running one job in our integration suite. (Merging to main in tigerdata-app will still trigger the entire test suite including mflux_ci like normal)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -304,9 +304,8 @@ workflows:
   version: 2
   build_accept_deploy:
     when:
-      or:
-        - equal: ["<< pipeline.parameters.GHA_Event >>", ""]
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -333,3 +332,9 @@ workflows:
             branches:
               only:
                 - main
+
+  integration_tests:
```
JaymeeH, line 336: CircleCI will now only run the `mflux_ci` job when triggered by the back-end (tigerdata-config)

bess (approved):
Good thinking!

PR #2639, opened 2026-04-28
Moving svg into a vue component.

carolyncole:
Cleans up the File Browser template and allows for the icon to be re-used

Changes the color of the icon to match figma and extracts it so it is not inline in the file browser.

## Original Screen
Screenshot 2026-04-28 at 2 01 11 PM

## Figma display
Screenshot 2026-04-28 at 2 01 05 PM

## New Screen
Screenshot 2026-04-28 at 2 00 09 PM


leefaisonr (approved):
Thanks @carolyncole !

PR #2642, opened 2026-04-28
Add hover effect to copy vue component

carolyncole:
moved frames and copy icons into vue
remove props passing around urls


bess (approved):
Great idea to split the SVG out, that's so much better. Thank you!

Issue #2641, opened 2026-04-28
If there are 100 files in the folder we should not display the file warning


carolyncole:
In /tigerdata/RDSS/Query/CProject/n_00100 there are exactly 100 item in the folder.  We should Not be displaying the limit as the folder only reaches the limit, it does not have more

carolyncole:
We discussed this in standup and determined not to put in a separate fix.  Hopefully when we upgrade mediaflux next there will be a solution https://support.arcitecta.com/hc/en-us/requests/9128

PR #2644, opened 2026-04-29
Connect the warning message to the medaiflux complete status during each folder load

carolyncole:
should allow us to see if we got exactly 100 back and that was all there were, but there is a bug in mediaflux 
This PR is not trying to address that bug

refs #2641

### DIFF
--- a/app/javascript/vue_components/file_browser.vue
+++ b/app/javascript/vue_components/file_browser.vue
@@ -12,7 +12,7 @@
       <copy-path class="col" :path="displayedPath"> </copy-path>
     </div>
 
-    <div v-if="displayedFiles.length >= fileDisplayLimit" class="preview-limit-frame">
+    <div v-if="displayWarning" class="preview-limit-frame">
       <div class="preview-limit-warning">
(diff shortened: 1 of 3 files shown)


app/javascript/vue_components/file_browser.vue
```diff
@@ -12,7 +12,7 @@
       <copy-path class="col" :path="displayedPath"> </copy-path>
     </div>
 
-    <div v-if="displayedFiles.length >= fileDisplayLimit" class="preview-limit-frame">
```
carolyncole, line 15: Notice how this was '>=' which meant if the work had 100 items exactly we would always show the message.  The new code still has the bug, but it is at a mediaflux level.

spec/system/project_spec.rb
```diff
@@ -130,7 +131,6 @@
         end
 
         it "displays an empty folder indicator" do
-          visit project_path(approved_project)
```
carolyncole, line 133: There was no reason to visit approved project without checking anything

bess (approved):

PR #2685, opened 2026-05-14
Small CSS fix and typo correction for the wizard tooltip

kelynch:
Some minor fixes I noticed while troubleshooting on another issue


leefaisonr (approved):
Thanks @kelynch

Issue #2675, opened 2026-05-12
Display users in the details panel in the standard way


matthewjchandler:
### Current Display

Currently, the display of users in the details panel does not match the standard display in other places, such as in the basic details at the top of the project page.

<img width="1168" height="1332" alt="Image" src="https://github.com/user-attachments/assets/684baec4-62f7-4f40-b4e5-49e766580dd1" />

### Expected Display

User values should display as "Firstname Lastname (netid)" where netid is a badge, following [Figma designs](https://www.figma.com/design/QthyYyx8IHYzUTGcHf0rdy/TigerData-UI-TM-DP?node-id=5794-5148&m=dev).

### Acceptance Criteria

- [x] In the details panel, display users in the standard way
  - [x] Instead of "(princeton:netid)" included in the name slot, show the netid in a gray badge (see [Figma](https://www.figma.com/design/QthyYyx8IHYzUTGcHf0rdy/TigerData-UI-TM-DP?node-id=5794-5148&m=dev)).
  - [x] If the domain is different from "princeton:" include it in parentheses with the uid in the name slot (no badge)
  - [x] If a given item is otherwise generated by the system, display the user value as "System Generated" in the name slot (no badge)
- [x] Modify the Mediaflux asset in `app/models/asset.rb` in Mediaflux::Asset; right now it is a single string, we would prefer a hash with the creator name and the UID separate

kelynch:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RDSS-61a4f1a12a399b001730f65a/issues/gh/pulibrary/tigerdata-app/2675?planning-poker) with Zenhub @carolyncole @JaymeeH @jrgriffiniii @leefaisonr

PR #2691, opened 2026-05-14
Display users in the details panel the standard way

JaymeeH:
ref #2675

### DIFF
--- a/app/javascript/vue_components/project_item.vue
+++ b/app/javascript/vue_components/project_item.vue
@@ -77,22 +77,29 @@
         </div>
         <div class="inline-container">
           <header class="fw-semibold" data-attribute-name="createdBy">Created By</header>
-          <template v-if="parseCreatedBy(displayedObject.created_by)">
+          <template v-if="systemGenerated()">
             <div class="createdByContainer">
               <p class="project-file-attribute font-monospace" data-attribute-name="createdBy">
-                {{ displayName }}
+                System Generated
+              </p>
+            </div>
+          </template>
+          <template v-else-if="princetonUserGenerated()">
+            <div class="createdByContainer">
+              <p class="project-file-attribute font-monospace" data-attribute-name="createdBy">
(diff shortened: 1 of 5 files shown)


app/javascript/vue_components/project_item.vue
```diff
@@ -96,6 +111,21 @@ const hiddenRoot = ref(props.hiddenRoot);
 const displayedPath = ref(props.currentObject.path.replace(hiddenRoot.value, ''));
 const displayedObject = ref(props.currentObject);
 
+const standardUser = ref(false);
+const displayName = ref('');
+const netid = ref('');
+
+function parseCreatedBy(createdBy) {
+  if (createdBy.includes('(')) {
+    displayName.value = createdBy.split('(')[0].trim();
```
carolyncole: @JaymeeH This code may be fragile.  I wonder if it would be better to pass the parts out in the Mediaflux Asset.  I squashed them together in [this PR](https://github.com/pulibrary/tigerdata-app/commit/7e5ff5d0dd8c115534b4aa1b5024facd056e0bbb#diff-adc310b43002d303d666db82feb7c5eecc262620beb4e5ec916241bd554e812cR62-R68) when sending them as separate elements would have made this work easier. If you [change this line](https://github.com/pulibrary/tigerdata-app/blob/7e5ff5d0dd8c115534b4aa1b5024facd056e0bbb/app/presenters/project_file_show_presenter.rb#L41) to pass multiple values instead of the one then no parsing would be necessary.

carolyncole (dismissed):
If you change the values being pass to the UI, you would not need as much logic and could handle dropping the princeton: pretty easily/

carolyncole (approved):
Thanks @JaymeeH This looks great! I deployed it to staging and there is only one small mishap I think.  But this is a giant step forward 

Image

JaymeeH:
Display for System Generated assets:
Screenshot 2026-05-28 at 11 35 35 AM

PR #2699, opened 2026-05-20
Upgrades for rspec and chromedriver

bess:
* Un-pin rspec so we can upgrade
* Set chromedriver to latest in devbox so we have a better chance at upgrading whenever chrome makes a new release. Hopefully they release chrome and chromedriver at around the same time?


leefaisonr (approved):
Thanks @bess

PR #2700, opened 2026-05-20
Upgrade rubocop

bess:
We have previously been using [bixby](https://github.com/samvera/bixby), the samvera style rules, for style checking. However, that project is no longer maintained, and we have fallen behind on recent rubocop updates.


app/controllers/new_project_requests_controller.rb
```diff
@@ -47,20 +45,18 @@ def approve
...
   end
-  # rubocop:enable Metrics/MethodLength
-  # rubocop:enable Metrics/AbcSize
-  # rubocop:enable Metrics/PerceivedComplexity
-  # rubocop:enable Metrics/CyclomaticComplexity
 
-  def destroy
+      # rubocop:enable Metrics/MethodLength
+      # rubocop:enable Metrics/AbcSize
+      def destroy
```
carolyncole, line 51: Th indentation here looks wrong.

PR #2718, opened 2026-06-02
Keep binstubs in version control

bess:
This addresses a warning that appeared in CircleCI that our binstubs needed regenerating. This might be because of some updates I was working on. It's apparently a best practice to keep the binstubs in version control. From the Rails docs:

```
When you install a gem whose executable you want to use in your app,
generate it and add it to source control:

  bundle binstubs some-gem-name
  git add bin/new-executable
```

More from the error message:
```
If you already have Rails binstubs in source control, you might be
inadvertently overwriting them during deployment by using bundle install
with the --binstubs option.

If your application was created prior to Rails 4, here's how to upgrade:

  bundle config --delete bin    # Turn off Bundler's stub generator
  rails app:update:bin          # Use the new Rails executables
  git add bin                   # Add bin/ to source control
```

I ran that recommended process and this PR is the result.


carolyncole (approved):
It deployed correctly.  Why not...

PR #2744, opened 2026-07-06
Added documentation on access tokens coming from CAS

leefaisonr:


bess (changes requested):
@leefaisonr This is a really good start! Would you mind please reviewing the Yard documentation and formatting this to match? 

I would not expect to see blank lines in a comment for a method. 
It would also be good to know what kind of object is returned by this method. 

Here are some docs that might be helpful: https://github.com/lsegal/yard/blob/main/docs/GettingStarted.md

PR #2752, opened 2026-07-07
Retry NewProjectRequest.last in rspec tests

bess:
Sometimes tests are failing b/c of a lag in the database. This change will retry the failing database query until it succeeds.


carolyncole (approved):
Thanks @bess

PR #2824, opened 2026-07-31
Pulling in the lando chrome concept from figgy

carolyncole:
Allows test to run locally without needing a chromedriver for the latest chrome that is installed on my machine

Advances https://github.com/pulibrary/tigerdata-app/issues/2754


bess (approved):
Thanks, @carolyncole

